PR #4, opened 2021-08-31
OAPEN Workflow and bug fixes to ONIX workflow

rhosking:
Ported pull request from observatory-platform

Adds the new OAPEN workflow, which will combine the IRUS-UK, and public metrics with the OAPEN metadata. (MEL-217 + MEL-159)

It also adds a range of fixes across the OAEBU workflows:

- corrects matching for countries in google analytics (MEL-371)
- changes irus-uk to use the correct country field not city locations to stop over counting (MEL-391)
- adds UCL metrics for UCL press (MEL-212)
- deduplication of metrics from Google Analytics, IRUS-UK, Google Books, UCL Press and JSTOR (MEL-387, MEL-386, MEL-379)
- ensures that duplicate ONIX product records don't throw off the metrics (MEL-380, MEL-379)
- Adds work and work_family IDs to the exported tables (where it makes sense) (MEL-385)
- Adds titles to the unmatched ISBNs (MEL-384)

### DIFF
--- a/oaebu_workflows/workflows/oapen_workflow.py
+++ b/oaebu_workflows/workflows/oapen_workflow.py
@@ -25,6 +25,7 @@ from airflow.models import Variable
 from airflow.sensors.external_task import ExternalTaskSensor
 from oaebu_workflows.config import sql_folder
 
+from oaebu_workflows.workflows.onix_workflow import make_table_id
 from observatory.platform.workflows.workflow import Workflow
 from observatory.platform.utils.gc_utils import (
     bigquery_sharded_table_id,
@@ -63,6 +64,7 @@ class OapenWorkflowRelease:
         :return: None.
         """
         pass
+
 
 class OapenWorkflow(Workflow):
     """
@@ -209,7 +211,6 @@ class OapenWorkflow(Workflow):
 
         release.cleanup()
 
-
     def create_onix_formatted_metadata_output_tasks(
         self,
         release: OapenWorkflowRelease,
@@ -276,12 +277,13 @@ class OapenWorkflow(Workflow):
         table_id = bigquery_sharded_table_id(output_table, release_date)
 
         # Identify latest Book release from the Academic Observatory
-        public_book_release_date = select_table_shard_dates(
+        public_book_table_id = make_table_id(
             project_id=self.ao_gcp_project_id,
             dataset_id=self.public_book_metadata_dataset_id,
             table_id=self.public_book_metadata_table_id,
             end_date=release.release_date,
-        )[0]
+            sharded=True,
+        )
 
         google_analytics_table_id = "empty_google_analytics"
         google_books_sales_table_id = "empty_google_books_sales"
@@ -307,7 +309,7 @@ class OapenWorkflow(Workflow):
             jstor_institution_table_id=jstor_institution_table_id,
             oapen_table_id=oapen_table_id,
             ucl_table_id=ucl_table_id,
-            public_book_release_date=public_book_release_date,
+            public_book_tabel_id=public_book_table_id,
         )
 
         create_bigquery_dataset(project_id=project_id, dataset_id=output_dataset, location=data_location)
--- a/oaebu_workflows/workflows/oapen_workflow.py
+++ b/oaebu_workflows/workflows/oapen_workflow.py
@@ -309,7 +309,7 @@ class OapenWorkflow(Workflow):
             jstor_institution_table_id=jstor_institution_table_id,
             oapen_table_id=oapen_table_id,
             ucl_table_id=ucl_table_id,
-            public_book_tabel_id=public_book_table_id,
+            public_book_tabel_id=f"{self.ao_gcp_project_id}.{self.public_book_metadata_dataset_id}.{public_book_table_id}",
         )
 
         create_bigquery_dataset(project_id=project_id, dataset_id=output_dataset, location=data_location)
--- a/oaebu_workflows/workflows/oapen_workflow.py
+++ b/oaebu_workflows/workflows/oapen_workflow.py
@@ -1,0 +1,568 @@
+# Copyright 2020 Curtin University
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#
+# Author: Richard Hosking
+
+import os
+import shutil
+from functools import partial, update_wrapper
+from pathlib import Path
+from typing import List, Optional
+
+import pendulum
+from airflow.exceptions import AirflowException
+from airflow.models import Variable
+from airflow.sensors.external_task import ExternalTaskSensor
+from google.cloud.bigquery import SourceFormat
+from observatory.dags.config import sql_folder
+
+from observatory.platform.workflows.workflow import AbstractRelease, Workflow
+from observatory.platform.utils.gc_utils import (
+    bigquery_sharded_table_id,
+    create_bigquery_dataset,
+    create_bigquery_table_from_query,
+    run_bigquery_query,
+    select_table_shard_dates,
+    copy_bigquery_table
+)
+from observatory.platform.utils.jinja2_utils import render_template
+from observatory.platform.utils.airflow_utils import AirflowVars
+from observatory.platform.utils.workflow_utils import (
+    bq_load_shard_v2,
+    table_ids_from_path,
+    make_dag_id,
+)
+
+
+class OapenWorkflowRelease(AbstractRelease):
+    """
+    Release information for OapenWorkflow.
+    """
+
+    def __init__(
+        self,
+        *,
+        dag_id: str,
+        release_date: pendulum.DateTime,
+        gcp_project_id: str,
+        ao_gcp_project_id: str = "academic-observatory",
+        oapen_metadata_dataset_id: str = "oapen",
+        oapen_metadata_table_id: str = "metadata",
+        public_book_metadata_dataset_id: str = "observatory",
+        public_book_metadata_table_id: str = "book",
+        irus_uk_dag_id_prefix: str = "oapen_irus_uk",
+        irus_uk_dataset_id: str = "oapen",
+        irus_uk_table_id: str = "oapen_irus_uk",
+        oaebu_dataset: str = "oaebu",
+        oaebu_onix_dataset: str = "oapen_onix",
+        oaebu_intermediate_dataset: str = "oaebu_intermediate",
+        oaebu_elastic_dataset: str = "data_export",
+        dataset_location: str = "us",
+        dataset_description: str = "Oapen workflow tables",
+    ):
+        """
+        :param dag_id: DAG ID.
+        :param release_date: The release date. It's the current execution date.
+        :param oapen_release_date: the OAPEN release date.
+        :param gcp_project_id: GCP Project ID.
+        :param ao_gcp_project_id: GCP project ID for the Academic Observatory.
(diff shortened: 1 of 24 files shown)


oaebu_workflows/workflows/oapen_workflow.py
```diff
@@ -0,0 +1,559 @@
...
+
+    new_table_id = table_id
+    if sharded:
+        table_date = select_table_shard_dates(
+            project_id=project_id,
+            dataset_id=dataset_id,
+            table_id=table_id,
+            end_date=end_date,
+        )[0]
+        new_table_id = f"{table_id}{table_date.strftime('%Y%m%d')}"
```
jdddog, line 195: This function is a duplicate of the function in onix_workflow.py module and can just be imported from there.

oaebu_workflows/workflows/oapen_workflow.py
```diff
@@ -0,0 +1,559 @@
...
+    return new_table_id
+
+
+class OapenWorkflow(Workflow):
+    """
+    Workflow for processing the OAPEN metadata and IRUS-UK metrics data
+    """
+
+    DAG_ID_PREFIX = "oapen_workflow"
+    ORG_NAME = "OAPEN"
```
jdddog, line 206: We've been calling it "OAPEN Press" in the API

oaebu_workflows/workflows/tests/test_oapen_workflow.py
```diff
@@ -0,0 +1,290 @@
...
+
+        self.data_location = os.getenv("TEST_GCP_DATA_LOCATION")
+
+        self.ao_gcp_project_id = "academic-observatory"
+        self.oapen_metadata_dataset_id = "oapen"
+        self.oapen_metadata_table_id = "metadata"
+        self.public_book_metadata_dataset_id = "observatory"
+        self.public_book_metadata_table_id = "book"
+
+        self.org_name = "OAPEN"
```
jdddog, line 135:
```suggestion
        self.org_name = OapenWorkflow.ORG_NAME
```

oaebu_workflows/workflows/oapen_workflow.py
```diff
@@ -0,0 +1,559 @@
...
+        self.add_task(
+            self.create_onix_formatted_metadata_output_tasks,
+            task_id="create_onix_formatted_metadata_output_tasks"
+        )
+
+        # Copy IRUS-UK data and add release date
+        self.add_task(
+            self.copy_irus_uk_release,
+            task_id="copy_irus_uk_release"
+        )
```
jdddog, line 265: Why does IRUS UK need to be copied? Can't it just be queried from wherever it is?

oaebu_workflows/workflows/oapen_workflow.py
```diff
@@ -0,0 +1,559 @@
...
+    create_bigquery_dataset,
+    create_bigquery_table_from_query,
+    select_table_shard_dates,
+    copy_bigquery_table
+)
+from observatory.platform.utils.jinja2_utils import render_template
+from observatory.platform.utils.airflow_utils import AirflowVars
+from observatory.platform.utils.workflow_utils import make_dag_id
+
+class OapenWorkflowRelease(AbstractRelease):
```
jdddog, line 40: So that you can remove all of the unused methods, you can just not subclass AbstractRelease, it will still work. Perhaps in the type hints in the Workflow template we should let release be Any type and then the type hints will work too.

jdddog (commented):
Hey Richard, it looks good to me overall, just a few small suggestions and questions

oaebu_workflows/workflows/oapen_workflow.py
rhosking: thanks, yeah makes sense. Will change that

oaebu_workflows/workflows/oapen_workflow.py
rhosking: I don't think it is technically a press, though I will check on that. It was wondering I have been wondering what the appropriate term would be

oaebu_workflows/workflows/oapen_workflow.py
rhosking: yeah, that was what I wanted to do originally. I will take another look at it with fresh eyes. Essentially, by reusing the sql.jinja file from the onix workflow, it was expecting the table name and location to confirm to a particular structure. I took the approach of trying to not have to change the onix workflow as much as possible

oaebu_workflows/workflows/oapen_workflow.py
aroelo: I think it would be better to not use the AbstractRelease class directly, I thought that the Release class was developed for this purpose, it has all the required methods/properties that you need to be able to use the Workflow class and the implemented methods.

Also, I think that all properties of the OapenWorkflowRelease (except for the release date) would probably be better in the OapenWorkflow class, as none of these are related to a specific release. They are just passed on to the release class and then extracted in the workflow class again where they are used.

oaebu_workflows/workflows/oapen_workflow.py
```diff
@@ -0,0 +1,559 @@
...
+            self.dag_id = make_dag_id(self.DAG_ID_PREFIX, self.ORG_NAME)
+
+        if airflow_vars is None:
+            airflow_vars = [
+                AirflowVars.DATA_PATH,
+                AirflowVars.PROJECT_ID,
+                AirflowVars.DATA_LOCATION,
+                AirflowVars.DOWNLOAD_BUCKET,
+                AirflowVars.TRANSFORM_BUCKET,
+            ]
```
aroelo, line 240: Are these variables required for the workflow? If so then it would be good to add the 'check_dependencies' task that is already implemented, if they are not needed this bit can probably be left out.

oaebu_workflows/workflows/oapen_workflow.py
jdddog: We are already using OAPEN Press though, so it would be better to use that for consistency

oaebu_workflows/workflows/oapen_workflow.py
jdddog: > I think it would be better to not use the AbstractRelease class directly, I thought that the Release class was developed for this purpose, it has all the required methods/properties that you need to be able to use the Workflow class and the implemented methods.

Some of the properties and methods in the AbstractRelease and Release class aren't really needed for all types of workflows.

> Also, I think that all properties of the OapenWorkflowRelease (except for the release date) would probably be better in the OapenWorkflow class, as none of these are related to a specific release. They are just passed on to the release class and then extracted in the workflow class again where they are used.

Yeah that sounds good, also gcp_project_id should be included.

oaebu_workflows/workflows/oapen_workflow.py
```diff
@@ -0,0 +1,525 @@
...
+    create_bigquery_dataset,
+    create_bigquery_table_from_query,
+    select_table_shard_dates,
+    copy_bigquery_table
+)
+from observatory.platform.utils.jinja2_utils import render_template
+from observatory.platform.utils.airflow_utils import AirflowVars
+from observatory.platform.utils.workflow_utils import make_dag_id
+
+class OapenWorkflowRelease(AbstractRelease):
```
jdddog, line 40:
```suggestion
class OapenWorkflowRelease:
```

oaebu_workflows/workflows/oapen_workflow.py
```diff
@@ -0,0 +1,525 @@
...
+        """Not used.
+        :return: Empty string.
+        """
+        return str()
+
+    def cleanup(self):
+        """Delete all files and folders associated with this release.
+        :return: None.
+        """
+        pass
```
jdddog, line 121: Can remove these

oaebu_workflows/workflows/oapen_workflow.py
```diff
@@ -0,0 +1,525 @@
...
+
+class OapenWorkflow(Workflow):
+    """
+    Workflow for processing the OAPEN metadata and IRUS-UK metrics data
+    """
+
+    DAG_ID_PREFIX = "oapen_workflow"
+    ORG_NAME = "OAPEN"
```
jdddog, line 129: The ORG_NAME needs to be changed to "OAPEN Press" otherwise the sensor will hang forever

jdddog (changes requested):

jdddog:
The code needs to be reformatted with Black

oaebu_workflows/workflows/oapen_workflow.py
```diff
@@ -0,0 +1,525 @@
...
+
+        # Initialise Telesecope base class
+        super().__init__(
+            dag_id=self.dag_id, start_date=start_date, schedule_interval=schedule_interval, catchup=catchup, airflow_vars=airflow_vars
+        )
+
+        # Wait for irus_uk workflow to finish
+        ext_dag_id = make_dag_id(irus_uk_dag_id_prefix, self.ORG_NAME)
+        sensor = ExternalTaskSensor(task_id=f"{ext_dag_id}_sensor", external_dag_id=ext_dag_id, mode="reschedule")
+        self.add_sensor(sensor)
```
jdddog, line 218: There need to be two sensors that wait for the following DAGs to finish:
* oapen_irus_uk_oapen_press
* oapen_metadata

This also means that ORG_NAME needs to be changed to "OAPEN Press".

jdddog (changes requested):

oaebu_workflows/workflows/oapen_workflow.py
rhosking: Take a look at the updates I made, I think they are right, but sensors aren't something I have worked with much, so feedback would be useful

oaebu_workflows/workflows/oapen_workflow.py
```diff
@@ -0,0 +1,466 @@
...
+            ucl=False,
+            onix_workflow=False,
+            onix_workflow_dataset="",
+            google_analytics_dataset="",
+            google_books_dataset="",
+            jstor_dataset="",
+            oapen_dataset=self.irus_uk_dataset_id,
+            ucl_dataset="",
+            public_book_release_date=public_book_release_date,
+        )
```
jdddog, line 333: The template `create_book_products.sql.jinja2` can be simplified so that you don't need to copy the `oapen_irus_uk` dataset and also don't need to specify so many variables for the template.

In the Jinja2 template you can check if a variable is defined or not:
```jinja2
{% if variable is not defined %}
```
So instead of having the variables `google_analytics=False, google_books=False, jstor=False, oapen=True, ucl=False, onix_workflow=False` you can just check if the table_id variable is defined or not.

Then, also change the template to supply the whole table_id. We might need to change some of our terminology to be consistent with BigQuery:
* table name: jstor or jstorYYYYMMDD
* table id: project_id.dataset_id.table_name.

So instead of this:
```sql
jstor_institution_metrics as (
    SELECT
        eISBN as ISBN13, release_date, ARRAY_AGG(STRUCT(Institution, Total_Item_Requests)) as metrics
    FROM {% if jstor %} `{{ project_id }}.{{ dataset_id }}.{{ jstor_dataset }}_jstor_institution_matched{{ release_date.strftime('%Y%m%d') }}` {% else %} empty_jstor_institution {% endif %} as jstor_institution
    GROUP BY eISBN, release_date
),
```

Can do this:
```sql
jstor_institution_metrics as (
    SELECT
        eISBN as ISBN13, release_date, ARRAY_AGG(STRUCT(Institution, Total_Item_Requests)) as metrics
    FROM {% if jstor_table_id is defined  %} `{{ jstor_table_id }}` {% else %} empty_jstor_institution {% endif %} as jstor_institution
    GROUP BY eISBN, release_date
),
```

Also, some of the variables passed to the template are unused, e.g. ucl, onix_workflow, onix_workflow_dataset etc, so would be good to double check them.

jdddog (changes requested):
Hey Richard, the changes look good. I've left some comments about simplifying some of the template code.

Also, can you change black to use a line length of 120 and reformat the changed files. In PyCharm in external tools settings, change the "Arguments" field to `"$FilePath$" --line-length 120`

oaebu_workflows/workflows/oapen_workflow.py
rhosking: I went a step further, and removed the logic from the jinja template, and did that in python. It was a bit of a pain using the logic for the full table_id betwen the onix and oapen workflows without doing this

jdddog (approved):

PR #9, opened 2021-09-01
Add TelescopeTypes to identifiers.py

aroelo:
Add TelescopeTypes class to `utils/identifiers.py` file.
I chose to add this class to a separate file rather than `config.py`, so that a new telescope type can easily be appended with the `generate workflow` command.
See also https://github.com/The-Academic-Observatory/observatory-platform/pull/485


jdddog (commented):
Thanks Aniek. I think you could just put the file in `oaebu_workflows/identifiers.py` without the utils folder.

aroelo:
> Thanks Aniek. I think you could just put the file in `oaebu_workflows/identifiers.py` without the utils folder.

Sure, I will move the file and rebase as well :-)

jdddog (approved):

PR #11, opened 2021-09-02
Fix oapen metadata for airflow2

tuanchien:


jdddog (approved):
Thanks for adding the tests and fixing it for Airflow 2 Tuan

PR #22, opened 2021-10-06
Update Google Books to handle multiple accounts for 1 publisher

aroelo:
So far publishers had 1 Google Books account per publisher, however some publishers have multiple Google Books account.
This PR adds functionality to handle this use case.

- The Google Book reports are in csv format and loaded into a list of records at the 'transform' step. When there are multiple accounts, these reports (of the same type and date) are merged together in the same list of records.
So there will be separate files for each account in the download bucket, but after transforming there will be just 1 file. The transformed data is then loaded into BQ as usual.

- To distinguish the files of different accounts a 'file_suffix' should be used. The file_suffix is part of the filename of the files on the SFTP server (where the reports are pulled from in the telescope). It would be useful to map the file_suffix to the Google Books account name, so you can easily tell which report maps to which account.
The documentation and Selenium script to create the proper file name is updated in this PR as well.

- For the Telescope instance created by our API, a list of file_suffixes should then be passed on as the telescope 'extra' object. So if your files on the sftp server for a specific publisher are:
  - GoogleBooksTrafficReport_foo2018_01.csv
  - GoogleBooksTrafficReport_bar2018_01.csv
  
  Then the telescope extra object would be
```python
{"file_suffixes": ["foo", "bar"]}
```

The telescope extra object can be None, so it does not need to be changed for most publishers which use just 1 Google Books account.

- When there are reports available on the SFTP server for e.g. 2 out of 3 accounts (of a specific date and type), none of the accounts will be processed in that release. The telescope knows how many reports to expect by the file_suffixes that are passed on. Only when there is a report for each account, these reports will be processed.


aroelo:
> For incomplete account reports, is the objective to make the user rectify missing records, or just do the best you can with the available information?
> 
> * For forcing user action it might even be better to throw an exception to get them to deal with it.
> * For "doing the best we can", perhaps it's better to process even an incomplete number of reports, and give the warning.  In this scenario there is a secondary question of what happens when the user wants to add the missing reports later.  Do they have to delete an older bq partition and manually rerun the dagrun for a specific period?

Thanks for the review @tuanchien!

When there are reports missing for an account, the objective would be to alert the user so they will download the reports and upload them to the sftp server.

Just some more info about the reports, there are 2 report types (Traffic and Sales) for each account. Initially, there was one account per publisher, but now there is a publisher that has 3 different accounts.
So with 'missing' reports, I mean that there are some reports available on the sftp server for the same report type and the same date, but e.g. only for 2 of the 3 accounts.
If there are no reports (of the same type and date) for any of the 3 accounts I don't think there should be any warning/action.
In general, the telescope does not keep track of which reports 'should' be there, because it is a manual job it is the user's responsibility to upload the right reports.

I don't think throwing an exception is appropriate, because you would still like to process all the other reports that are available (from different dates and/or different types), with the exception the telescope will just stop completely.

I don't think processing incomplete reports is a good idea, if with incomplete you mean that there are reports of the same type and date, but not for all accounts.
If with incomplete you mean that e.g. reports of Traffic are all there, but not those for Sales, then these will already still be processed in the current set-up, it is only for missing reports in terms of missing for one of the accounts.

The reason I think it is not a good idea to process the incomplete reports is because it is not possible to see in the table (and therefore elasticsearch) that information is missing for one of the accounts. This might lead to numbers not matching up when creating the visualisation and the people making those visualisations don't have easily access to check whether the data is complete, because it can only be found in the airflow logs.
And additionally, like you mentioned if you do want to add the missing data you would have to manually rerun that dag run. Not sure if you would have to delete the partition or if it would get overwritten.

tuanchien:
Thanks for the additional context @aroelo.  I guess if that is your intended user interaction then that seems sensible.

PR #24, opened 2021-10-06
INF-444 Create additional dag instance for exporting public data

rhosking:
This also requires corresponding updates to airflow conns and additions to the Elastic Workflow classes


rhosking:
closing as no longer the approach taken

PR #32, opened 2021-11-04
changed month metrics filter

rhosking:
The export scripts were being overzealous in dropping rows


jdddog (approved):
Thanks Richard, looks good!

PR #39, opened 2021-11-15
Fix DOAB telescope

aroelo:
The DAG run initially failed, because the CSV file that was downloaded was empty.
I'm not sure how this happened, when I downloaded it today the CSV was populated.

To prevent this or at least keep an eye on it I've added a check to verify that the CSV is not empty in the download step.
If it is empty hopefully it will be fixed automatically when the task is retried.

When I reran the tasks from 'download' onwards, the csv was now downloaded correctly, but the DAG run still fails because it seems that the schema also changed, so I've updated the schema in this PR as well.


jdddog (approved):
Thanks Aniek, looks good!

jdddog (approved):

PR #40, opened 2021-11-15
Fix OAPEN Metadata telescope

aroelo:
Updated missing field to the schema


jdddog (approved):
Thanks Aniek, it looks good!

PR #48, opened 2021-11-22
Update OAPEN IRUS UK telescope

aroelo:
Updated the telescope to separately pass on the publisher name that is used for counter v4 platform and the publisher uuid that is used for counter v5 platform.

For the publishers used so far, the publisher name could be used to look up the publisher uuid, this was done inside the telescope.
However, as it turns out some publishers have different publisher names/IDs for v4 and v5.
The best way to get the publisher UUID is to look this up from the results returned by the v5 OAPEN API (described in the updated docs in this PR)

This is why it is necessary to specify the publisher name and uuid separately.
This also means that the existing Telescope instances will have to be updated with our Observatory API.

See https://github.com/The-Academic-Observatory/oapen-irus-uk-cloud-function/pull/9 for related cloud function PR

### DIFF
--- a/docs/telescopes/oapen_irus_uk.md
+++ b/docs/telescopes/oapen_irus_uk.md
@@ -40,21 +40,57 @@ The corresponding table created in BigQuery is `oapen.oapen_irus_ukYYYYMMDD`.
 ```
 
 ## Telescope object 'extra'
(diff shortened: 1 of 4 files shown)


jdddog (approved):
Thanks Aniek, looks good

PR #49, opened 2021-12-01
Fix JSTOR email filtering

aroelo:
- Replace single quote with double quote to fix subject filtering
- Filter on sender as well


jdddog (approved):
Thanks Aniek!

PR #56, opened 2022-02-01
Change DAG id for public umich export

jdddog:
Change DAG id for public umich export from university_of_michigan_press to university_of_michigan_press_public as it was a duplicate of the private export DAG id.


aroelo (approved):
Nice, thanks for updating

PR #63, opened 2022-03-09
Use new countries table, rather than the two previous tables.

jdddog:
This PR refactors the OnixWorkflow and the OapenWorkflow to use the new COKI country table. Two new fields are added to 
export_book_metrics_country indexes, including: country_iso_name and country_wikipedia_name, so that drop down lists can be matched against either a country's ISO name or a country's Wikipedia article name.


aroelo (approved):
Do you think it is necessary to test this in a develop GCP/elastic environment similar to some of the google analytics changes?
The changes are quite straight forward so maybe not, but if you are interested I can show you how I set this up.

jdddog:
> Do you think it is necessary to test this in a develop GCP/elastic environment similar to some of the google analytics changes? The changes are quite straight forward so maybe not, but if you are interested I can show you how I set this up.

Thanks Aniek, I'll have a think about it and let you know.

PR #66, opened 2022-03-21
Fix oapen metadata 'invalid date' issue

aroelo:
One of the entries in the latest oapen metadata release has a `dc.date.issued` value of "0215".

This is likely a typo and should be "2015". Regardless of the cause, this value is currently formatted into a date format using `pendulum.parse(date).to_date_string()`.
The result of this is "215-01-01" on linux and "0215-01-01" on mac (see https://bugs.python.org/issue13305 explaining difference).

The date format accepted by BigQuery is `'YYYY-[M]M-[D]D'`, so "215-01-01" leads to the error that we see in Airflow:
```
Invalid date: '215-01-01' Field: issued; Value: 215-01-01
```

To fix this I have updated
`pendulum.parse(date).to_date_string()`
to
`pendulum.parse(date).to_date_string().rjust(10, '0')` so the date string will always include a 4 digit year by appending zeros to the start of the string.

### DIFF
--- a/oaebu_workflows/workflows/oapen_metadata_telescope.py
+++ b/oaebu_workflows/workflows/oapen_metadata_telescope.py
@@ -181,7 +181,7 @@ class OapenMetadataTelescope(StreamTelescope):
             schema_folder,
             schema_prefix=schema_prefix,
             airflow_vars=airflow_vars,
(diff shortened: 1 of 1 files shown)


jdddog (approved):
Thanks Aniek, it looks good.

PR #68, opened 2022-03-25
Use country.google_analytics_name for matching with Google Analytics

jdddog:
Google Analytics uses their own variation of country names, which are used as a part of a join in oaebu_workflows/database/sql/export_book_metrics_country.sql.jinja2, so a new field has been added to the country table that contains the names used in Google Analytics (rather than being in a separate table).


aroelo (approved):
Thanks @jdddog

PR #69, opened 2022-03-27
MEL-399 Added BISAC content

niamhq:


aroelo (approved):
Thanks @niamhq, looks good!

PR #80, opened 2022-03-31
Mel 778/update overview and new schema

kathrynnapier:
New schema files with the google analytics and country fields added, and minor text changes to oaebu project docs to be consistent with data inventory doc.


aroelo (approved):
Thanks @kathrynnapier, looks good :-)

PR #84, opened 2022-04-12
Update CITATION.cff

bechandcock:
Updated the citation title in the .cff file from 'OAeBU Workflows' to 'Book Usage Data Workflows'


aroelo (approved):
Thanks @bechandcock

PR #86, opened 2022-04-14
Update Google Books schedule interval

aroelo:
Update schedule from monthly to weekly. 
The google books reports need to be updated manually to the SFTP server, because of this manual process they might be updated at any point during the month.
It will be better to run the telescope every week to check if there are new reports available, so the kibana dashboards can be updated during the weekend with the latest google books data.


jdddog (approved):
Thanks Aniek, it looks good.

PR #89, opened 2022-04-21
INF-384: Use 2nd gen cloud function

aroelo:
Use second generation cloud function, primarily because this has a longer maximum time out.

### DIFF
--- a/oaebu_workflows/workflows/oapen_irus_uk_telescope.py
+++ b/oaebu_workflows/workflows/oapen_irus_uk_telescope.py
@@ -221,8 +221,10 @@ class OapenIrusUkTelescope(OrganisationTelescope):
         "https://github.com/The-Academic-Observatory/oapen-irus-uk-cloud-function/releases/"
         "download/v1.1.6/oapen-irus-uk-cloud-function.zip"
     )  # URL to the zipped source code of the cloud function
-    FUNCTION_MD5_HASH = "e9b350a31a43314a994404631dcfae3a"  # MD5 hash of the zipped source code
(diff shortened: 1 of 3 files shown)


oaebu_workflows/workflows/oapen_irus_uk_telescope.py
```diff
@@ -426,15 +435,19 @@ def create_cloud_function(
...
-        "ingressSettings": "ALLOW_ALL",
-        "sourceArchiveUrl": os.path.join("gs://" + source_bucket, blob_name),
-        "httpsTrigger": {"securityLevel": "SECURE_ALWAYS"},
+        "buildConfig": {
+            "runtime": "python39",
+            "entryPoint": "download",
+            "source": {"storageSource": {"bucket": source_bucket, "object": blob_name}},
+        },
+        "serviceConfig": {
+            "timeoutSeconds": 900,
```
jdddog, line 446: Is the maximum timeout 60 minutes? It would be good to define it in a constant and link to the documentation about what the true maximum is.

jdddog (changes requested):
It looks good Aniek, I would just put a little bit of documentation in to help with understanding how to change the maximum timeout (see comment).

oaebu_workflows/workflows/oapen_irus_uk_telescope.py
aroelo: Yeah it is 60 minutes now, thanks for the comment. I have added some more info!

jdddog (approved):
Thanks Aniek, looks good!

PR #91, opened 2022-05-11
Fix oapen metadata dc.date.issued field

aroelo:


jdddog (approved):
Thanks Aniek, it looks good to me.

PR #93, opened 2022-05-31
Use DatasetTypeId class

aroelo:
Update telescopes that use the StreamTelescope class to use the dataset_type_id


jdddog (approved):
Thanks Aniek, looks good to me.

PR #94, opened 2022-06-07
Inf 406/add scripts to seed api db

jdddog:


jdddog (commented):
Hey Tuan, it looks good, however this branch doesn't seed the database properly, there are problems with the Onix Workflow, there could be others. 

For this workflow, could you please:
* Run it in your local environment and make sure that all of the workflows seed.
* Update the info classes to use the oaebu_workflows.api_type_ids constants.
* Update the branch to use the updated seeding pattern, see the observatory-platform and academic-observatory-workflows branches for comments about what I've changed.
* Double check that the output is the same as the migration script that we ran previously.

PR #95, opened 2022-06-10
BAD-142: ANU google analytics update

keegansmith21:
Adding the page views metric to the google analytics telescope. Updated the SQL query that created the export tables so that google analytics data is not longer ignored in some circumstances.


oaebu_workflows/database/schema/google_analytics_1900-01-01.json
```diff
@@ -97,6 +97,74 @@
...
+          {
+            "mode": "NULLABLE",
+            "name": "value",
+            "type": "INTEGER",
+            "description": "Number of page views."
+          }
+        ]
+      }
+    ]
+  },
```
aroelo, line 167: This schema is not used actively right now, because we only pull in Google Analytics data for ANU Press. But if we do update this it would probably be better to keep this file unchanged and add a completely new schema `oaebu_workflows/database/schema/google_analytics_2022-06-13.json`, similar to the `oaebu_workflows/database/schema/anu_press_google_analytics_2022-06-13.json` one.

aroelo (changes requested):
Great work @keegansmith21! It definitely was a bit tricky working with the Google Analytics API and inspecting at each point of the process where the numbers do or don't match up.

I just added one small comment on the google analytics BigQuery schema.

oaebu_workflows/database/schema/google_analytics_1900-01-01.json
keegansmith21: Thanks for the suggestion @aroelo, I've updated the files. I believe that the workflow will choose the latest suitable one based on the datestamps. Can you conform that this is correct?

jdddog: Yeah it finds the latest schema based on the date.

jdddog (approved):

PR #98, opened 2022-06-17
Add dag tags

tuanchien:
Requires https://github.com/The-Academic-Observatory/observatory-platform/pull/571

### DIFF
--- a/oaebu_workflows/dag_tag.py
+++ b/oaebu_workflows/dag_tag.py
@@ -1,0 +1,4 @@
+class Tag:
+    """DAG tag."""
+
+    oaebu = "oaebu"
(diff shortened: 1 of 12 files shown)


jdddog (changes requested):
Thanks Tuan, these changes look good.

We should also add the `oaebu` TAG to all of the oaebu elastic_import DAGs. It will need to be a parameter passed to the workflow in the DAG file.

tuanchien:
Elastic/vm tagging changes made.  Requires merging https://github.com/The-Academic-Observatory/observatory-platform/pull/572 first

jdddog (approved):
Thanks Tuan, it looks good to me.

PR #99, opened 2022-06-27
Parallelise some onix workflow tasks

tuanchien:
Not clear if actually faster to parallelise.  Slightly slower in tests, but that could just be timing variations.  The proof is in the pudding.  Might be worth testing 1 run using production configuration to see if it's worth using.


jdddog (approved):
Thanks Tuan, it looks good to me.

PR #100, opened 2022-07-11
Fixed the google analytics schema

keegansmith21:
The google analytics schemas currently break as the 2022 version attempts to find the 'page_views' from the previous versions. Instead, we can just update the 1900 version to get the page_views and propagate the changes forward.


jdddog (approved):
Thanks Keegan, it looks good to me.

PR #102, opened 2022-07-25
Bad 214 oapen onix unittest table updates

keegansmith21:
Removed dependency on the observatory-unit-testing project for the oapen workflow unit test and created mock data to replace the previously used real data. The test also no longer queries data from the academic-observatory dataset.

### DIFF
--- a/oaebu_workflows/workflows/tests/test_oapen_workflow.py
+++ b/oaebu_workflows/workflows/tests/test_oapen_workflow.py
@@ -363,47 +363,48 @@ class TestOapenWorkflowFunctional(ObservatoryTestCase):
                     )
                 )
 
+                # Create tables and test conditions
+                release_suffix = release_date.strftime("%Y%m%d")
+
                 # Format OAPEN Metadata like ONIX to enable the next steps
                 ti = env.run_task(workflow.create_onix_formatted_metadata_output_tasks.__name__)
                 self.assertEqual(expected_state, ti.state)
+                onix_table_id = f"{self.gcp_project_id}.{oaebu_onix_dataset_id}.onix{release_suffix}"
+                self.assert_table_integrity(onix_table_id, expected_rows=2)
+                self.assert_table_content(
+                    onix_table_id, load_jsonl(test_fixtures_folder("oapen_workflow", "expected_onix.jsonl"))
+                )
 
-                # Create oaebu output tables
+                # Create book product table
                 ti = env.run_task(workflow.create_oaebu_book_product_table.__name__)
                 self.assertEqual(expected_state, ti.state)
+                book_product_table_id = f"{self.gcp_project_id}.{oaebu_output_dataset_id}.book_product{release_suffix}"
+                self.assert_table_integrity(book_product_table_id, expected_rows=2)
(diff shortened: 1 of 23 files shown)


jdddog (changes requested):
Thanks Keegan, it's much better using the mocked data now.

I just realised that there are three additional external tables in, which should also be mocked as well, the sql templates and the tables that they are used in are:
- export_book_subject_thema_metrics.jinja2
  - oaebu-public-data.oaebu_reference.thema_lookup
- export_book_subject_bic_metrics.jinja2
  - oaebu-public-data.oaebu_reference.bic_lookup
- export_book_subject_bisac_metrics.jinja2
  - oaebu-public-data.oaebu_reference.bisac_lookup

The above three tables are quite simple, should just be able to export them as CSV or JSON Lines from BigQuery, save to Git LFS and load them like you did the others.

The unit tests that checks the table content could be more thorough (I know you didn't write them!) Some ideas:
* Check that each table exists with `self.assert_table_integrity`. All of the export tables, onix and book_product tables.
* Then for the onix and book_product table, make a JSON lines file with their expected output, read it in and compare it to the actual output. You can use the `self.assert_table_content` test for this.

keegansmith21:
Currently blocked by The-Academic-Observatory/observatory-platform#575

oaebu_workflows/workflows/tests/test_oapen_workflow.py
```diff
@@ -399,3 +426,20 @@ def test_run_workflow_tests(self):
...
+            onix_table_id = f"{self.gcp_project_id}.{oaebu_onix_dataset_id}.onix{release_suffix}"
+            book_product_table_id = f"{self.gcp_project_id}.{oaebu_output_dataset_id}.book_product{release_suffix}"
+            self.assert_table_integrity(onix_table_id)
+            self.assert_table_integrity(book_product_table_id)
+            self.assert_table_content(
+                onix_table_id, load_jsonl(test_fixtures_folder("oapen_workflow", "expected_onix.jsonl"))
+            )
+            self.assert_table_content(
+                book_product_table_id, load_jsonl(test_fixtures_folder("oapen_workflow", "expected_book_product.jsonl"))
+            )
```
jdddog, line 445: It would be good to put the tests below each task that they related to. So asserting the ONIX table content should go below `create_onix_formatted_metadata_output_tasks`, then the book product table content check below `create_oaebu_book_product_table` and the export table checks after the export tables are created. Perhaps also add the expected number of rows to `assert_table_integrity`.

You could also remove the "Check records in book_product and book_product_list match" tests by adding the expected number of rows to `assert_table_integrity` where the export tables are checked.

jdddog (changes requested):
Thanks Keegan, that's looking good, just a few small comments.

oaebu_workflows/workflows/tests/test_oapen_workflow.py
keegansmith21: Thanks Jamie. I've made some updates and it's ready for a re-review.

jdddog (approved):
Thanks Keegan, the changes look great.

PR #105, opened 2022-08-19
Bad 210 wp2 standardised data ingest pipelines workflows

keegansmith21:
Integrated Crossref events and metadata retrieval into the Onix workflow to eliminate the dependency on The Academic Observatory repository.

The Onix Workflow will now query the crossref metadtata and events APIs using the ISBNs from the Onix table. It will then create its own "book" table and use this to create the book_products table for each publisher.

A similar process is done for the Oapen workflow, which also requires Crossref data.


oaebu_workflows/workflows/tests/test_onix_workflow.py
```diff
@@ -2098,14 +2344,55 @@ def run_telescope_tests(self, *, org_name: str, include_google_analytics: bool =
...
+                # Assertions
+                self.assertEqual(expected_state, ti.state)
+                self.assert_table_content(
+                    f"{gcp_project_id}.{oaebu_crossref_dataset_id}.crossref_metadata{release_suffix}",
+                    load_jsonl(test_fixtures_folder("onix_workflow", "crossref_metadata.jsonl")),
+                )
+
+                # Load crossref event table into bigquery
+                events_vcr = vcr.VCR(record_mode="none", before_record_request=vcr_ignore_condition)
+                with events_vcr.use_cassette(self.events_cassette):
```
jdddog, line 2381: You should just be able to write:
```python
with vcr.use_cassette(self.events_cassette, record_mode="none", before_record_request=vcr_ignore_condition):
```

You would use vcr.VCR when you wanted to use the same settings with many different vcr.use_cassette calls.

oaebu_workflows/workflows/onix_workflow.py
```diff
@@ -218,43 +243,107 @@ class OnixWorkflow(Workflow):
...
+    CROSSREF_EDITED_EVENT_URL_TEMPLATE = (
+        "https://api.eventdata.crossref.org/v1/events/edited?mailto={mailto}"
+        "&from-updated-date={start_date}&until-updated-date={end_date}&rows=1000"
+        "&obj-id={doi}"
+    )
+    CROSSREF_DELETED_EVENT_URL_TEMPLATE = (
+        "https://api.eventdata.crossref.org/v1/events/deleted?mailto={mailto}"
+        "&from-updated-date={start_date}&until-updated-date={end_date}&rows=1000"
+        "&obj-id={doi}"
+    )
```
jdddog, line 300: These can all be moved out of the `OnixWorkflow` class and just be constants in this module.

What if there are more than 1000 rows?

jdddog (changes requested):
Hey Keegan, it's looking good, I just have a few comments and questions (below). Have you run it end to end, locally for a partner?

oaebu_workflows/workflows/onix_workflow.py
keegansmith21: The 1000 rows is a limit enforced by the API. For the events we use a cursor to iterate through pages of events. I purposely neglected to do this with the metadata as I did not expect it to be possible to have over 1000 DOIs for a single ISBN. After chatting with Cameron it appears that this is a possibility, so I'll build it into the workflow.

keegansmith21:
@jdddog  - Comments have been addressed and ready for re-review

jdddog (approved):

PR #107, opened 2022-08-24
Remove old buckets and datasets with same prefix

alexmassen-hane:
Added functions from [PR](https://github.com/The-Academic-Observatory/observatory-platform/pull/578) to make a common prefix for the buckets and datasets and will delete them if they are older than 7 days. The prefix is unique to each test and will not delete buckets or datasets created or being used by other tests. 

Unit tests will fail until [PR](https://github.com/The-Academic-Observatory/observatory-platform/pull/578) is merged.


alexmassen-hane:
This branch can be closed because we have changed when the check for old datasets/buckets are run. They are now in the ObservatoryEnvironment class and is run every time there is a .create() used.

PR #109, opened 2022-08-25
Bad 241 - UCL Discovery data collection - Added ISBN10s

alexmassen-hane:
Added column of "isbn" to the data pull because they are supposed to contain ISBN10s but sometimes also contain ISBN13s and vice versa. By adding this change, we can make a list of effected eprint IDs with dirty ISBNs that we can send to the publisher and request them to fix the errors.


jdddog (approved):
Thanks Alex, it looks good.

PR #111, opened 2022-09-05
INF-540: Updated JSTOR telescope to ignore new fields in the schema

keegansmith21:
The JSTOR telescope has run into issues with new data. The data cannot be loaded into the table as there are now new fields. We would prefer new fields to be ignored by default so that this doesn't happen.


jdddog (approved):
Thanks Keegan, it looks good!

PR #113, opened 2022-09-07
BAD-276: Fulcrum Telescope Development

keegansmith21:
## Fulcrum Telescope

Similar to the Thoth Telescope (#119), Fulcrum forgoes the use of the telescope templates in favor of a more modular approach. 

The telescope can be compared to the oapen_irus_uk telescope as it hits the same API (using a different platform number - 235). However, because no sensitive IP information is used, it is much simpler to operate.

The telescope workflow runs once a month for every publisher that makes use of the Fulcrum platform. Each run will download its respective month's data, transform it and add it to a partitioned table in BigQuery.

Currently, only University of Michigan is using Fulcrum

### DIFF
--- a/oaebu_workflows/workflows/fulcrum_telescope.py
+++ b/oaebu_workflows/workflows/fulcrum_telescope.py
@@ -1,0 +1,409 @@
+# Copyright 2023 Curtin University
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Author: Keegan Smith
+
+import logging
+import os
+from typing import Dict, List, Optional, Tuple
+
+import pendulum
+from airflow.exceptions import AirflowException
+from airflow.hooks.base import BaseHook
+from google.cloud.bigquery import SourceFormat
+from google.cloud.bigquery.table import TimePartitioningType
+
+from oaebu_workflows.config import schema_folder as default_schema_folder
+from oaebu_workflows.dag_tag import Tag
+from observatory.platform.utils.airflow_utils import AirflowConns, make_workflow_folder
+from observatory.platform.utils.file_utils import list_to_jsonl_gz, list_files, load_jsonl, blob_name_from_path
+from observatory.platform.utils.url_utils import retry_get_url
+from observatory.platform.utils.config_utils import find_schema
+from observatory.platform.utils.gc_utils import upload_files_to_cloud_storage
+from observatory.platform.utils.workflow_utils import (
+    bq_load_partition,
+    table_ids_from_path,
+    make_release_date,
+    cleanup,
+)
+from observatory.platform.workflows.workflow import Release, Workflow
+
+IRUS_FULCRUM_ENDPOINT_TEMPLATE = (
+    "https://irus.jisc.ac.uk/api/v3/irus/reports/irus_ir/?platform=235"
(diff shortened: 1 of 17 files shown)


oaebu_workflows/workflows/fulcrum_telescope.py
```diff
@@ -0,0 +1,406 @@
...
+        "University of Michigan Bentley Library",
+        "University of Michigan Center for Chinese Studies",
+        "University of Michigan Center for Japanese Studies",
+        "University of Michigan Center for South Asia Studies",
+        "University of Michigan Center for South East Asian Studies",
+        "University of Michigan Department of Near Eastern Studies",
+        "University of Michigan Museum of Anthropological Archaeology",
+        "University of Michigan Press",
+    ]
+}
```
jdddog, line 61: When you update this on the new config file branch, it would be good if this org name to publisher mapping was supplied via an argument in kwargs, then the mapping can be configured without needing to change the code.

oaebu_workflows/workflows/fulcrum_telescope.py
```diff
@@ -0,0 +1,406 @@
...
+        dataset_id: str = "fulcrum",
+        merge_partition_field: str = "release_date",
+        schema_folder: str = default_schema_folder(),
+        source_format: str = SourceFormat.NEWLINE_DELIMITED_JSON,
+        airflow_conns: List = None,
+        catchup: bool = True,
+        download_file_name="fulcrum.jsonl.gz",
+        transform_file_name="fulcrum.jsonl.gz",
+        dataset_description: str = "IRUS Fulcrum Data Feed",
+        load_bigquery_table_kwargs={"ignore_unknown_values": True},
```
jdddog, line 98: Default keyword arguments of type dictionary or list set in this way are mutable, meaning that they can be changed. The implication of this is that if you modified this dictionary somewhere in your code, and this constructor (or function) was called again, the `load_bigquery_table_kwargs` would be set to the same object that you modified, not the value `{"ignore_unknown_values": True}`. 

So you have to jump through some hoops in Python to make this work properly. You need to set `load_bigquery_table_kwargs: Dict = None` and then if this is set to None, you assign the value `{"ignore_unknown_values": True}`. The effect is that when the user hasn't set a value for `load_bigquery_table_kwargs` the default value that you want will be used, and if they have set a value, then what they have set will be used.

oaebu_workflows/workflows/fulcrum_telescope.py
```diff
@@ -0,0 +1,406 @@
...
+        download_month = release.release_date.format("YYYY-MM")
+        totals_data, country_data = download_fulcrum_month_data(download_month, requestor_id)
+        totals_path = os.path.join(self.download_totals_folder, self.download_file_name)
+        country_path = os.path.join(self.download_country_folder, self.download_file_name)
+
+        if not totals_data or not country_data:
+            raise AirflowException(f"Data not available for supplied release date: {self.download_month}")
+
+        list_to_jsonl_gz(totals_path, totals_data)
+        list_to_jsonl_gz(country_path, country_data)
```
jdddog, line 201: In the updated branch, these have changed to save_jsonl_gz: https://github.com/The-Academic-Observatory/observatory-platform/blob/feature/workflow-config-file/observatory-platform/observatory/platform/files.py#L219

jdddog (changes requested):
Thanks Keegan, the documentation and code looks good. Just a small suggestion for now and some comments for when you update it on the new config file branch.

oaebu_workflows/workflows/fulcrum_telescope.py
keegansmith21: Ahhh, I remember hearing about this once but not understanding. Thank you for the explanation, I'll update this. I still find it strange that Python behaves this way for lists/dicts.

oaebu_workflows/workflows/fulcrum_telescope.py
keegansmith21: Yeah I agree that would be much nicer. I'll make the change in the upcoming branch

oaebu_workflows/workflows/fulcrum_telescope.py
keegansmith21: Yea I'll have to do an overhaul for the upcoming changes

jdddog (approved):

PR #114, opened 2022-09-14
INF-550: Updated JSTOR transform for new report format

keegansmith21:
JSTOR report format has changed in the following ways:

- Removal of the Usage Month column
- Renaming of Total Item Requests to Reporting Period Total
- Addition of a date column for each month that the report spans

These changes break our workflow, so I have adjusted the telescope's transform function such that the original headings are retained


jdddog (approved):
Thanks for fixing this Keegan, it looks good to me.

PR #115, opened 2022-10-19
INF-562: JSTOR schedule update

keegansmith21:
JSTOR reports are scheduled to be emailed every month on the 1st. Our workflows also run on the 1st of every month, which means that occasionally the workflows miss the reports.


jdddog (approved):

PR #116, opened 2022-11-02
BAD-308: schema and BQ description updates

keegansmith21:
The schema files can differ from the BQ tables since they are not used in their creation. This has led to the divergence of the BQ table schema and the documentation schema.

This update uses the PR from The-Academic-Observatory/observatory-platform#589 to make use of the schema.json files when creating the BQ tables. This accomplishes two desirable things:

- Synchronicity of the documentation with reality, where the schema.json files act as a singular source of truth
- Descriptions from the schema.json files will be pushed to the BQ tables 

As a result, the workflows will now issue an error when a SQL query creates a table that does not fit the expected schema. You may still neglect to supply a schema file, which will not raise any issues and upload the data with an implied schema.


keegansmith21:
Reopening this PR because I've changed the names of some of the schema files (removed dates). 

This changed is blocked by The-Academic-Observatory/observatory-platform#589

keegansmith21:
All tests passing locally

jdddog (changes requested):
Hey Keegan, looks good, I think we can change a few of the schemas.

Are these used?
* book_2021-11-25.json
* book_product_2021-11-25.json
* book_product_2022-03-31.json

These are partitioned tables so can have the dates removed:
* doab_2021-01-01.json
* ucl_discovery_2008-01-01.json
* oapen_metadata_2018-05-14.json

These can have the dates removed as they are output tables:
* onix_aggregate_metrics_2021-11-25.json
* onix_invalid_isbn_2022-03-27.json
* platform_invalid_isbn_2022-03-27.json
* platform_unmatched_isbn_2022-03-27.json

Can we just keep the latest schema for the various oaebu_public_data and oaebu_publisher_book_product schemas and remove the dates because they are output tables?

jdddog (approved):
Thanks Keegan, looks good.

PR #117, opened 2022-11-03
Removed the loading of the OAEBU service account credentials

keegansmith21:
UCL telescope test loads the OAEBU service account credentials for some reason. It's not used at all, so I've removed it.


jdddog (approved):
Thanks Keegan, looks good!

PR #118, opened 2022-11-25
Bad 257 OAPEN ONIX Integration

keegansmith21:
This is a PR for integrating OAPEN's metadata into the onix telescope and onix workflow


oaebu_workflows/workflows/oapen_metadata_workflow.py
```diff
@@ -0,0 +1,394 @@
...
+    positive_offset = -1
+    line = all_lines[line_number - 1]
+    while not "</Product>" in line:
+        positive_offset += 1
+        line = all_lines[line_number + positive_offset]
+
+    if (line_number - negative_offset) < 0:
+        raise AirflowException(f"Product out of bounds for given line number {line_number}")
+
+    return (line_number - negative_offset, line_number + positive_offset)
```
jdddog, line 347: I'm not sure how feasible this is, so just disregard it if it seems like too much work.

It would be cleaner if you could use the internals of `validate_onix` as it traverses the XML with the library it uses, e.g. lxml, and then use the same library to delete the invalid elements, write the valid XML out to file and create a new XML tree, add the invalid products and write them to a file. Maybe `validate_onix` isn't design to make a task like this easy though.

oaebu_workflows/workflows/oapen_metadata_workflow.py
```diff
@@ -0,0 +1,394 @@
...
+
+
+class OapenMetadataWorkflow(StreamTelescope):
+    """Oapen Metadata Workflow"""
+
+    METADATA_URL = "https://library.oapen.org/download-export?format=onix"
+    DAG_ID = "oapen_metadata_workflow"
+    SFTP_UPLOAD_DIR = "/telescopes/onix/oapen_press/upload"
+
+    def __init__(
```
jdddog, line 80: I think it would be easier to keep this named as the Oapen Metadata Telescope, as it kind of is that, we are just changing what format the data is output in and where it is saved, then we don't need to make any changes to anything related to seeding apart from removing the oapen_workflow and don't need to update the production database, aside from deleting things related to oapen_workflow.

oaebu_workflows/workflows/onix_telescope.py
```diff
@@ -288,6 +291,16 @@ def __init__(
         )
 
         # self.organisation = organisation
+        if self.organisation_name == "OAPEN Press":
```
jdddog, line 294: It would be better to not add a hard coded sensor into the OnixTelescope. Instead, you could use the pattern below:

```python
class MyWorkflow(Workflow):
    def __init__(
        self,
        *,
        sensor_dag_ids: List[str] = None,
        ...
    ):
        ...
        self.sensor_dag_ids = sensor_dag_ids
        if sensor_dag_ids is None:
            self.sensor_dag_ids = []

        # Add a list of ExternalTask sensors to wait for other DAGs to finish
        for dag_id in self.sensor_dag_ids:
            self.add_sensor(
                ExternalTaskSensor(
                    task_id=f"{dag_id}_sensor",
                    external_dag_id=dag_id,
                    mode="reschedule",
                )
            )

```

You would need to store the sensor_dag_ids in the extras field of the workflow in the API, as a JSON list, then in dags/onix_workflow.py you would need to grab the list of sensor_dag_ids from the API with workflow.extras.get("sensor_dag_ids") and parse the JSON list.

jdddog (changes requested):
Thanks for the PR Keegan, it looks good, just some comments about keeping the previous name of the telescope so that we don't have to make too many changes when deploying it and making the sensors for the ONIX Telescope able to be specified via a parameter in the constructor.

oaebu_workflows/workflows/onix_workflow.py
```diff
@@ -1855,7 +1833,7 @@ def isbns_from_onix(project_id: str, onix_dataset_id: str, onix_table_id: str) -
     isbn_rows = run_bigquery_query(sql)
     isbns = [row["ISBN13"] for row in isbn_rows]
     isbns = list(filter(None, isbns))  # Remove null values
-    return isbns
+    return isbns[:200]
```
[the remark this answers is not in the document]
keegansmith21: Good pickup, I was meant to change this back post-test

oaebu_workflows/workflows/oapen_metadata_workflow.py
keegansmith21: Yeah I'm aware that this is a pretty subpar solution. It turns out handling XML files is just kinda hard.

I didn't look into doing what you suggest here when writing this, but I've had a play around now and I think it's probably more trouble than its worth. The onixcheck library mostly just wraps the lmxl library and adds a few schemas to validate against that conform to ONIX standards. The lxml library is great for validating, but the issue is that it can't do it iteratively as far as I can see. It also doesn't have any way to store/return the xml elements that contain the invalid components; it can only tell you the line in which the error occurs. I think this is closer to a limitation of the xml format than with the library itself though, since it'd have no way of knowing that you are specifically looking for the second-to top level element that contains each error.

oaebu_workflows/workflows/oapen_metadata_workflow.py
keegansmith21: Damn, wish I'd known this before changing everything! haha

But no problem. I only called it a workflow because it's removing a few things about the data it's gathering.

oaebu_workflows/workflows/oapen_metadata_workflow.py
jdddog: No worries!

jdddog (dismissed):

jdddog (approved):

PR #119, opened 2022-11-25
BAD-372: Thoth Telescope

keegansmith21:
# Thoth

This is a PR for the creation of the Thoth telescope into OAEBU workflows.

This also updates the API seeding to create the Open Book Publishers organisation and the relevant workflows.

## Design

The design of this telescope varies from others, as we wanted to minimise the layers of abstraction (inheritance) and the reliance on the release class. Comments and suggestions are very welcome!

The Thoth release class holds only the release date and DAG ID (this too can be removed using dynamic airflow tasks post version upgrade #112). The rest of the information normally held in the release class has been delegated to the telescope class.

The telescope inherits from the base 'Workflow' class rather than any of the specific telescope classes. This removes one layer of inheritance in its design, making its functionality and operation more transparent. This also means that some functions that would normally be called from the specific telescope classes (ie. streamtelescope, snapshottelescope) have been recreated as standalone functions (get_data_path(), blob_name_from_path(), make_workflow_folder()). Currently these functions are stored in thoth_telescope.py, but these should be moved to a more generic utility file as more telescopes make use of them.

Recreating some of the core functions of the telescope allows for the redesign of the folder hierarchy. The updated hierarchy can be deduced from [this line](https://github.com/The-Academic-Observatory/oaebu-workflows/pull/119/files#diff-9bd08955dc1abcb10519d5f53eb247208b1fe879a5dfd5a0826866ac76985ae3R297). The result should look like this:
data_path/thoth_onix_org/thoth_onix_org_2023_01_01/download
data_path/thoth_onix_org/thoth_onix_org_2023_01_01/transform

## Workflow
The workflow for Thoth is quite simple:
1. Download publisher metadata (ONIX file)
2. Transform the metadata in much the same way as the ONIX telescope
3. Upload the ONIX feed to Bigquery

Since this is effectively an alternate way of retrieving an ONIX feed, it uses some of the ONIX labelling (including having the same dataset ID). However, the DAG ID and download files are prefixed with "thoth" to avoid confusion. The workflow also shares the same 'onix.json' schema file, as we would expect this to be no different to the onix telescope's.

### DIFF
--- a/oaebu_workflows/workflows/thoth_telescope.py
+++ b/oaebu_workflows/workflows/thoth_telescope.py
@@ -1,0 +1,359 @@
+# Copyright 2023 Curtin University
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Author: Keegan Smith
+
+import os
+import shutil
+from typing import List
+import logging
+
+import pendulum
+from airflow.exceptions import AirflowException
+from airflow.models import Variable
+from google.cloud.bigquery import SourceFormat
+from airflow.secrets.environment_variables import EnvironmentVariablesBackend
+
+from observatory.platform.utils.airflow_utils import AirflowVars
+from oaebu_workflows.config import schema_folder as default_schema_folder
+from oaebu_workflows.workflows.onix_telescope import parse_onix
+from observatory.platform.utils.config_utils import find_schema
+from observatory.platform.utils.url_utils import retry_session
+from observatory.platform.utils.file_utils import list_files
+from observatory.platform.utils.gc_utils import upload_files_to_cloud_storage
+from observatory.platform.utils.workflow_utils import (
+    SubFolder,
+    bq_load_shard,
+    table_ids_from_path,
+    make_release_date,
+    delete_old_xcoms,
+)
+from observatory.platform.workflows.workflow import Workflow, Release
+
+THOTH_URL = "{host_name}/specifications/{format_specification}/publisher/{publisher_id}"
+DEFAULT_FORMAT_SPECIFICATION = "onix_3.0::oapen"
+DEFAULT_HOST_NAME = "https://export.thoth.pub"
+
+
+class ThothRelease(Release):
+    def __init__(
+        self,
+        *,
+        dag_id: str,
+        release_date: pendulum.DateTime,
+    ):
+        """Construct a ThothRelease.
+
+        :param dag_id: the DAG id.
+        :param release_date: the release date.
+        """
+        self.release_date = release_date
+        release_id = f'{dag_id}_{self.release_date.format("YYYY_MM_DD")}'
+        super().__init__(dag_id, release_id)
+
+
+class ThothTelescope(Workflow):
+    DAG_ID_PREFIX = "thoth_onix"
+
+    def __init__(
+        self,
+        *,
+        dag_id: str,
+        project_id: str,
+        download_bucket: str,
+        transform_bucket: str,
+        data_location: str,
+        publisher_id: str,
+        airflow_vars: List[str] = None,
+        start_date: pendulum.DateTime = pendulum.datetime(2022, 12, 1),
+        schedule_interval: str = "@weekly",
+        dataset_id: str = "onix",
+        schema_folder: str = default_schema_folder(),
+        source_format: str = SourceFormat.NEWLINE_DELIMITED_JSON,
+        catchup: bool = False,
+        workflow_id: int = None,
+        host_name: str = "https://export.thoth.pub",
+        format_specification: str = DEFAULT_FORMAT_SPECIFICATION,
+        download_file_name: str = "thoth_onix.xml",
+        transform_file_name: str = "thoth_onix.jsonl",
+        dataset_description: str = "Thoth ONIX Feed",
+    ):
+        """Construct an ThothOnixTelescope instance.
+
+        :param organisation_name: the organisation name.
+        :param project_id: the Google Cloud project id.
+        :param download_bucket: the Google Cloud download bucket.
+        :param transform_bucket: the Google Cloud transform bucket.
+        :param data_location: the location for the BigQuery dataset.
+        :param publisher_id: the publisher ID. Can be found using Thoth's GrapihQL API
+        :param airflow_vars: list of airflow variable dependencies, for each connection, it is checked if it exists in airflow.
+        :param dag_id: the id of the DAG, by default this is automatically generated based on the DAG_ID_PREFIX
+        and the organisation name.
+        :param start_date: the start date of the DAG.
+        :param schedule_interval: the schedule interval of the DAG.
+        :param dataset_id: the BigQuery dataset id.
+        :param schema_folder: the SQL schema path.
+        :param source_format: the format of the data to load into BigQuery.
+        :param catchup: whether to catchup the DAG or not.
+        :param workflow_id: api workflow id.
+        :param format_specification: The format to use when downloadin Thoth's metadata
+        :param download_file_name: The name of the file to write the ONIX data to
+        :param transform_file_name: The name of the file to write the transformed ONIX data to
+        :param dataset_description: The description to give to the BigQuery table
+        """
+
+        if airflow_vars is None:
+            airflow_vars = [
+                AirflowVars.DATA_PATH,
+                AirflowVars.PROJECT_ID,
+                AirflowVars.DATA_LOCATION,
+                AirflowVars.DOWNLOAD_BUCKET,
+                AirflowVars.TRANSFORM_BUCKET,
+            ]
+
+        # Cloud workspace settings
+        self.project_id = project_id
+        self.download_bucket = download_bucket
+        self.transform_bucket = transform_bucket
+
+        # Databse settings
+        self.dataset_description = dataset_description
+        self.dataset_id = dataset_id
+        self.data_location = data_location
+        self.source_format = source_format
+        self.schema_folder = schema_folder
+
+        # Thoth settings
+        self.publisher_id = publisher_id
+        self.host_name = host_name
+        self.format_specification = format_specification
+        self.download_file_name = download_file_name
+        self.transform_file_name = transform_file_name
+
+        # Initialise folders
+        self.download_folder = None
+        self.transform_folder = None
+
+        super().__init__(
+            dag_id=dag_id,
+            start_date=start_date,
+            schedule_interval=schedule_interval,
+            catchup=catchup,
+            airflow_vars=airflow_vars,
+            workflow_id=workflow_id,
+        )
+
+        # self.organisation = organisation
+        self.add_setup_task(self.check_dependencies)
+        self.add_task(self.download)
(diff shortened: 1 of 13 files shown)


keegansmith21:
@jdddog just a question about the design that would be good to have your input on when you're ready to review:
This currently works as a standalone telescope in that it'll pull data from Thoth, transform it and create a BQ table. There is another option for the telescope design that I did not consider until after creating it... We could make the workflow follow a similar pattern as with the OAPEN metadata telescope (#118). That is to say, we could download the onix XML (from Thoth) and upload it to the SFTP server's upload file. Then we could simply let the ONIX telescope pick up the file and process it as normal. This would have the advantage of confining the entirety of ONIX processing to a single telescope type. Please let me know what you think :smiley:

oaebu_workflows/workflows/thoth_telescope.py
```diff
@@ -0,0 +1,331 @@
...
+        self.add_task(self.upload_transformed)
+        self.add_task(self.bq_load)
+        self.add_task(self.cleanup)
+        self.add_task(self.add_new_dataset_releases)
+
+    def make_release(self, **kwargs) -> ThothRelease:
+        release_date = make_release_date(**kwargs)
+        release = ThothRelease(dag_id=self.dag_id, release_date=release_date)
+        self.download_folder = make_workflow_folder(self.dag_id, release_date, SubFolder.downloaded.value)
+        self.transform_folder = make_workflow_folder(self.dag_id, release_date, SubFolder.transformed.value)
```
jdddog, line 172: I would modify this a little bit, so that the folder created by `get_data_path(), dag_id, f"{dag_id}_{release_string}"` can be accessed separately, e.g. as self.workflow_folder.

Perhaps then `make_workflow_folder` could just create the path at `workflow_folder` and you can make the download_folder and transform_folder after this with os.path.join.

This is so that you can use `self.workflow_folder` to cleanup the whole workflow folder.

oaebu_workflows/workflows/thoth_telescope.py
```diff
@@ -0,0 +1,331 @@
...
+        download_onix(
+            publisher_id=self.publisher_id,
+            download_folder=self.download_folder,
+            format_spec=self.format_specification,
+        )
+
+    def upload_downloaded(self, release: ThothRelease, **kwargs) -> None:
+        """Upload the downloaded thoth onix XML to google cloud bucket
+
+        :raises AirflowException: Raised if there is not exactly 1 file in the donwload folder matching the expected name
```
jdddog, line 186: Add the remainder of the docstrings.

oaebu_workflows/workflows/thoth_telescope.py
```diff
@@ -0,0 +1,331 @@
...
+
+    def make_release(self, **kwargs) -> ThothRelease:
+        release_date = make_release_date(**kwargs)
+        release = ThothRelease(dag_id=self.dag_id, release_date=release_date)
+        self.download_folder = make_workflow_folder(self.dag_id, release_date, SubFolder.downloaded.value)
+        self.transform_folder = make_workflow_folder(self.dag_id, release_date, SubFolder.transformed.value)
+        return release
+
+    def download(self, release: ThothRelease, **kwargs) -> None:
+        """Task to download the ONIX release from Thoth."""
```
jdddog, line 176: Add the remainder of the docstrings.

oaebu_workflows/workflows/thoth_telescope.py
```diff
@@ -0,0 +1,331 @@
...
+        download_onix(
+            publisher_id=self.publisher_id,
+            download_folder=self.download_folder,
+            format_spec=self.format_specification,
+        )
```
jdddog, line 181:
```suggestion
        thoth_download_onix(
            host_name=self.host_name,
            publisher_id=self.publisher_id,
            download_folder=self.download_folder,
            format_spec=self.format_specification,
        )
```

oaebu_workflows/workflows/thoth_telescope.py
```diff
@@ -0,0 +1,331 @@
...
+            )
+
+        blob = blob_name_from_path(download_files[0])
+        success = upload_files_to_cloud_storage(self.download_bucket, [blob], download_files)
+        if not success:
+            raise AirflowException("Blob could not be uploaded to cloud storage")
+
+    def transform(self, release: ThothRelease, **kwargs) -> None:
+        """Task to transform the Thoth ONIX data"""
+        parse_onix(self.download_folder, self.transform_folder, make_org_id(self.organisation_name))
```
jdddog, line 201: The source variable of `parse_onix` is used to create a field called `COKI_ID`, which we don't use anywhere. We can remove the need for `make_org_id(self.organisation_name)` by removing this from the ONIX Parser.

These are the lines to remove in ONIX Parser, along with the associated `data_src` variables:
https://github.com/The-Academic-Observatory/onix-parser/blob/main/src/main/java/academy/observatory/app/OnixParser.java#L208-L209

oaebu_workflows/workflows/thoth_telescope.py
```diff
@@ -0,0 +1,331 @@
...
+        # Rename file
+        shutil.move(
+            os.path.join(self.transform_folder, "full.jsonl"),
+            os.path.join(self.transform_folder, self.transform_file_name),
+        )
+
+    def upload_transformed(self, release: ThothRelease, **kwargs) -> None:
+        """Upload the downloaded thoth onix .jsonl to google cloud bucket"""
```
jdddog, line 209: Add remaining docstrings.

oaebu_workflows/workflows/thoth_telescope.py
```diff
@@ -0,0 +1,331 @@
...
+            dataset_description=self.dataset_description,
+        )
+
+    def cleanup(self, release: ThothRelease, **kwargs) -> None:
+        """Delete all files and folders associated with this release."""
+        for path in [self.download_folder, self.transform_folder]:
+            try:
+                shutil.rmtree(path)
+            except FileNotFoundError as e:
+                logging.warning(f"No such file or directory {path}: {e}")
```
jdddog, line 253: It would be useful to move the cleanup functionality into a reusable function, as this will be reused across workflows, see below for function. Typically the cleanup should remove the workflow folder and also remove any XComs.

```suggestion
        """Delete all files and folders associated with this release."""
        
        cleanup(self.workflow_folder, self.dag_id, kwargs["execution_date"])
```

oaebu_workflows/workflows/thoth_telescope.py
```diff
@@ -0,0 +1,331 @@
...
+
```
jdddog, line 254: We can remove the entire workflow_folder rather than each subfolder.

```suggestion
def cleanup(workflow_folder, dag_id: str, execution_date):
    # Cleanup workflow folder
    try:
        shutil.rmtree(workflow_folder)
    except FileNotFoundError as e:
        logging.warning(f"No such file or directory {workflow_folder}: {e}")

    # Cleanup XComs
    delete_old_xcoms(dag_id=dag_id, execution_date=execution_date)
```

oaebu_workflows/workflows/thoth_telescope.py
```diff
@@ -0,0 +1,331 @@
...
+
+def download_onix(
+    publisher_id: str,
```
jdddog, line 257: If hostname can be altered, then better to let it be specified with an argument.

```suggestion
    *,
    host_name: str,
```

oaebu_workflows/workflows/thoth_telescope.py
```diff
@@ -0,0 +1,331 @@
...
+    download_folder: str,
+    download_filename: str = "thoth_onix.xml",
+    format_spec: str = "onix_3.0::oapen",
+    num_retries: int = 3,
+) -> None:
+    """Hits the Thoth API and requests the ONIX feed for a particular publisher.
+    Creates a file called onix.xml at the specified location
+
```
jdddog, line 265:
```suggestion

    :param host_name: the Thoth hostname.
```

oaebu_workflows/workflows/thoth_telescope.py
```diff
@@ -0,0 +1,331 @@
...
+    :param download_folder: The path of the download folder
+    :param download_filename: The name of the downloaded file
+    :param format_spec: The ONIX format to use. Options can be found with the /formats endpoint of the API
+    :param num_retries: The number of times to retry the download, given an unsuccessful return code
+    """
+    url = (
+        THOTH_BASE_URL
+        + THOTH_FORMAT_EXTENSION.format(format_specification=format_spec)
+        + THOTH_PUBLISHER_EXTENSION.format(publisher_id=publisher_id)
+    )
```
jdddog, line 276:
```suggestion
    url = THOTH_URL.format(host_name=host_name, format_specification=format_spec, publisher_id=publisher_id)
```

oaebu_workflows/workflows/thoth_telescope.py
```diff
@@ -0,0 +1,331 @@
...
+    """Return the path to this dag release's workflow folder. Will also create it if it doesn't exist
+
+    :param dag_id: The ID of the dag. This is used to find/create the workflow folder
+    :param release_date: The release date
+    :param folder_name: The name of a folder to create inside the workspace. e.g. 'download' or 'transform'
+    :return: the path of the workflow folder
+    """
+    release_string = release_date.format("YYYY_MM_DD")
+    path = os.path.join(get_data_path(), dag_id, f"{dag_id}_{release_string}", folder_name)
+    if not os.path.exists(path):
```
jdddog, line 298: I probably put this line before `if not os.path.exists(path):` but it is not necessary.

```suggestion
```

jdddog (changes requested):
Hey Keegan, the workflow looks good, it is much easier to understand having less inheritance. 

> This currently works as a standalone telescope in that it'll pull data from Thoth, transform it and create a BQ table. There is another option for the telescope design that I did not consider until after creating it... We could make the workflow follow a similar pattern as with the OAPEN metadata telescope (#118).

I prefer how you have made the workflow here, because it is a more straightforward way to get the data into BigQuery.

The main changes suggested are:
* `publisher_id` can be a parameter passed to the workflow and stored in the extra field of the workflow in the API.
* It would be good to remove `organisation_name`. It is inconsistent that this parameter
is used to create the `dataset_description` inside the workflow, and the `dag_id` from outside the workflow
in the dag file. I think it would be cleaner to remove it
entirely and set both the `dag_id` and `dataset_description` in the dag file. We can remove the need for the organisation name for `parse_onix`.
* By having a separate variable for the `workflow_folder`, we can make the cleanup simpler, and also put the cleanup functionality into a function.

oaebu_workflows/workflows/thoth_telescope.py
keegansmith21: What would you think of the following as a solution?:
```suggestion
def make_workflow_folder(dag_id: str, release_date: pendulum.DateTime, *subdirs: str) -> str:
    release_string = release_date.format("YYYY_MM_DD")
    path = os.path.join(get_data_path(), dag_id, f"{dag_id}_{release_string}", *subdirs)
    os.makedirs(path, exist_ok=True)
    return path
```

oaebu_workflows/workflows/thoth_telescope.py
keegansmith21: Sure! I've create a [PR](https://github.com/The-Academic-Observatory/onix-parser/pull/7) for the ONIX Parser

oaebu_workflows/workflows/thoth_telescope.py
jdddog: Yeah that looks good

keegansmith21:
@jdddog I've implemented the changes you've requested. Apart from those, I also made these changes:

- Added test for the parse_onix() function in the onix_telescope
- Removed date_format from onix_telescope and API seed as it does not appear to be used anywhere

jdddog (approved):
Hey Keegan, the changes look really good!

PR #122, opened 2022-12-15
BAD-412: Added view creation to onix workflow

keegansmith21:
Since we are using GCP's Looker, we need a set of statically named export tables that the dashboard can pull from. 

This PR creates a step at the end of the Onix Workflow to create a set of BigQuery views for each updated data export and QA table.


jdddog (approved):
Thanks Keegan, it looks good to me

PR #123, opened 2023-01-05
Oapen docs

keegansmith21:
The OAPEN metadata telescope needed a documentation update, as #118 has greatly altered the workflow of the telescope.
Also updated the csv schema generation to work with dateless json schemas, which are now common practice since #116


jdddog (approved):
Looks good Keegan

PR #124, opened 2023-01-05
BAD-314: CEU Press onboarding

keegansmith21:
We are onboarding CEU Press.
This involves updating the API seed to add the CEU organisation and relevant workflows and datasets.


jdddog (approved):
Thanks Keegan

PR #126, opened 2023-01-12
OAPEN metadata telescope type update

keegansmith21:
Updated the OAPEN metadata telescope type to a snapshot telescope


jdddog (approved):

PR #127, opened 2023-02-08
Onix keyword null ammendment

keegansmith21:
The Oapen onix workflow is running into issues with the book_product table creation. This appears to be because of the Subjects.SubjectHeadingText column of the Onix table - which is null-filled due to the absence of data from Oapen's XML metadata file. This PR slightly alters the SQL template that creates the book product table so that there are no issues with the SubjectHeadingText column being null-filled. 

Big thanks to @JulianTonti for helping with the solution.


jdddog (approved):
Thanks Keegan, looks good.

PR #128, opened 2023-02-10
Onix workflow's integration with onix telescope alternatives

keegansmith21:
This PR has become necessary because of the thoth telescope and related onboarding of OBP. Previously, there was no way for the onix workflow to look for an alternative onix telescope dag or table_id. 

~~I have added a field to the OnixWorkflow creation that allows it to use a different Onix telescope as a sensor. The source that it should use is provided to it via the observatory API. Additionally, a different table name can be used in place of the regular name (since thoth onix tables are named "thoth_onix" rather than "onix").~~
I have updated the API seed to list the ONIX telescopes as an 'oaebu' tagged workflow and updated the ONIX dataset type with the required 'extra' field. Now, the ONIX sensor no longer needs to be explicitly supplied in the ONIX workflow. Since OBP uses the thoth telescope for its ONIX feed, the ONIX OBP workflow will use this telescope as a sensor in place of a traditional ONIX telescope.

While making this, I also noticed that the Thoth transformed onix feed is in a different format to what we normally expect. I added a step in the transform process to fix this.


jdddog (changes requested):
Hey Keegan, I think a cleaner approach would be to make the ONIX datasets a data partner and to change the table name output by the Thoth telescope to "onix" so that there are less differences.

We should make it easier to specify data partners when we move to a config file.

keegansmith21:
Hi @jdddog, I've addressed your comments. Adding ONIX as a data partner to the API is a very clean solution. The only thing that needed attention was two portions of the ONIX workflow. These are the parts that create the intermediate and QA tables. It doesn't make sense for ONIX to have these tables, so they are explicitly excluded.

oaebu_workflows/workflows/onix_workflow.py
```diff
@@ -400,8 +399,7 @@ def __init__(
         )
 
         # Wait for data partner workflows to finish
-        partner_data_dag_prefix_ids = [OnixTelescope.DAG_ID_PREFIX]
-        partner_data_dag_prefix_ids.extend(list(set([partner.dag_id_prefix for partner in data_partners])))
+        partner_data_dag_prefix_ids = list(set([partner.dag_id_prefix for partner in data_partners]))
```
jdddog, line 402: Since the ONIX dataset is mandatory, maybe we should make an assertion to make sure that there is one here and raise an AirflowException if not? (To catch out when we don't configure it)

jdddog (commented):
Hey Keegan, the changes look good, just a small comment.

keegansmith21:
Thanks @jdddog. I've addressed your comment, which took a bit more effort than I expected for such a simple change. All of the difficulty of it was updating the tests to work by adding an ONIX dataset/workflow to the mock API. I've tried my best to keep it as clean as possible, as the onix workflow is quite extensive.

jdddog (approved):
Thanks Keegan, the changes look good!

PR #129, opened 2023-02-16
INF-591: Updated HTTP requests to use new retry function

keegansmith21:
This PR updates GET requests to use the retry_get_url function.

While doing this, I have also altered the default max_threads for onix_workflows and the crossref_events download task. This has become necessary as the oapen onix_workflow makes almost 190k http requests. Taking up to 24 hours to complete. The increase in threads used should decrease the time taken appreciably. The crossref events API limits requests to 15 per second, so I have limited the maximum threads for the download job to 15. Going beyond this number is counter-productive as a 429 is returned from crossref, prompting a short (but noticeable) wait before retrying the request.

### DIFF
--- a/oaebu_workflows/workflows/onix_workflow.py
+++ b/oaebu_workflows/workflows/onix_workflow.py
@@ -20,15 +20,15 @@ import shutil
 from datetime import timedelta
 from functools import partial, update_wrapper
 from pathlib import Path
-from typing import List, Optional
+from typing import List, Optional, Tuple
 import json
-from urllib3.exceptions import MaxRetryError
 import re
 
 from concurrent.futures import ThreadPoolExecutor, as_completed
 import pendulum
 from airflow.exceptions import AirflowException
 from google.cloud.bigquery import SourceFormat, Client
+from tenacity.wait import wait_exponential_jitter
 
 from oaebu_workflows.config import schema_folder as default_schema_folder
 from oaebu_workflows.config import sql_folder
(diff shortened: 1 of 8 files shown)


oaebu_workflows/workflows/onix_workflow.py
```diff
@@ -2105,18 +2111,15 @@ def download_crossref_event_url(url: str, i: int = 0):
     return events
 
 
-def download_crossref_page_events(url: str, headers: dict) -> tuple:
+def download_crossref_page_events(url: str, headers: dict) -> Tuple[str, int, int, List[dict]]:
```
jdddog, line 2114: If Crossref Events limits the number of calls per second, then it would be more robust to use a library that is able to limit the number of calls to a fixed number for a certain period.

An example is the ratelimit library which we use in the Scopus telescope: https://pypi.org/project/ratelimit/

```python
from ratelimit import limits, sleep_and_retry

@sleep_and_retry
@limits(calls=15, period=1)
def my_function():
    pass
```

Not sure how you would combine it with the code in `retry_session`, but maybe it would be enough to decorate `download_crossref_page_events`.

We should also update the CrossrefEventsTelescope.

jdddog (changes requested):
Thanks Keegan, it looks good. My only comment is that we could use a library to limit the number of calls per second to Crossref Events.

oaebu_workflows/workflows/onix_workflow.py
keegansmith21: I've made a small update implementing this change. It doesn't work perfectly (I think due to the threaded processes), but it reduces the number of 429 responses we get. Provided the threads are kept limited to 15, it's pretty good. I've implemented a similar solution to [The Academic Observatory](https://github.com/The-Academic-Observatory/academic-observatory-workflows/pull/157)

jdddog (approved):
Thanks Keegan, the changes look good to me.

PR #131, opened 2023-02-22
Updated filenames in dag to use telescope default

keegansmith21:
The onix table name derives from the file names used to create them. It is therefore imperative that the files are correctly named. The file names were recently updated on the telescope, but I forgot to update them as they are passed from the dag creation. I think that it is better to have the dag simply use the default of the telescope to avoid this in future.


jdddog (approved):
Thanks Keegan, looks good.

PR #133, opened 2023-02-24
BAD-490: Removed field selection from crossref metadata uri

keegansmith21:
It has been recommended to us that we refrain from using too many field selections when accessing the crossref metadata API as these are expensive for crossref. Instead, we can return all of the fields from the API and filter the ones we need.


jdddog (approved):

PR #135, opened 2023-02-27
Oapen metadata structure change and personname field

keegansmith21:
Added personname to oapen metadata

Also changed the structure of the telescope. It now parses the ONIX data and pushes it directly to bigquery rather than uploading it to the SFTP server, to be picked up by the OAPEN ONIX telescope. As such, the OAPEN ONIX telescope is no longer necessary.

I have added a logical statement in the onix workflow to properly create the sensor for this telescope. It's not the cleanest solution, but I believe it can be removed soon as we are moving to a configuration-file setup.


jdddog (approved):
Thanks Keegan, the PR looks good.

PR #136, opened 2023-03-03
INF-598: ONIX telescope keyword parsing

keegansmith21:
We are running issues when making the book_product table for many of the new telescopes. Each time it errors, it's because the _Subjects.SubjectHeadingText_ field is improperly populated for _Keywords_. The Keywords should not be a repeated field, but a single string where each keyword is separated by a semicolon. This is the only format that the SQL will reliably work with. This PR adds a step during the transform phase of the ONIX telescope to collapse repeated keywords into a single string and also attempts to change incorrect separators (commas, colons) to semicolons (as these also cause errors).


jdddog (approved):
Thanks Keegan, the changes look good to me.

PR #137, opened 2023-03-03
INF-604: Crossref pools for the ONIX workflow

keegansmith21:
This PR creates a pool for each of the crossref (metadata and event) APIs used by the ONIX workflow. This is necessary because many of the workflows can run at once, each of which will throttle themselves without knowledge of outside tasks.

The pool sizes were chosen based only on my experimentation. The slots that each task take up will be equal to the number of threads used for each process, thereby effectively limiting the total number of active threads across all tasks using the same pool.

Tested locally and appears to work as intended. Although it's difficult because running 5 tasks at once tanks my computer.


jdddog (approved):

PR #138, opened 2023-03-06
Fixes incorrect parameter in select_table_shard_dates function

jdddog:
Fixes incorrect parameter in select_table_shard_dates function


keegansmith21 (approved):
Thanks for this update Jamie!

PR #139, opened 2023-03-24
Updated telescope documentation index

keegansmith21:
I previously added the Thoth documentation as markdown but neglected to add the page to the index file. This PR fixes that and removes the unnecessary oapen_metadata schema.


jdddog (approved):
Thanks Keegan, looks good.

PR #140, opened 2023-03-28
Handle case: OAPEN Metadata returns only the XML header

keegansmith21:
Oapen will sometimes only return a header when requesting the metadata. No idea why. But it doesn't fall under any of the error cases in the download task. This PR causes an error to be raised if no XML "Product" is found in the downloaded content


jdddog (approved):

PR #141, opened 2023-03-31
Oapen metadata transform bucket consistency

keegansmith21:
The bucket name for bq_load was using the release bucket ID. This got past the tests... but we're refactoring them soon so I didn't bother adding this check to the test.


jdddog (approved):
Looks good Keegan

PR #142, opened 2023-03-31
Feature/workflow config file

keegansmith21:
# OAEBU Refactor

This PR is a refactor of the repository such that it is compatible with the upcoming [change to the Observatory Platform](https://github.com/The-Academic-Observatory/observatory-platform/pull/606).

## Purpose for Refactor
The primary purpose is to heavily reduce the reliance on the observatory API for the workflows. The API's only function now is to store dataset releases. Additionally, we wanted to reduce the reliance on inheritance for telescope and release creation and remove much of the release functionality. Now, the telescopes/workflows each inherit from the base Workflow class. Additional tools and utilities are available to fill the roles that the Stream/Snapshot/Organisation telescopes held. This change was necessary as it was restrictive to have to adhere to one of these three templates. There is also no concept of an _Organisation_ as this was found to be largely unnecessary.

## Release Class 
The release class for each telescope now inherits from either the PartitionRelease or SnapshotRelease depending on whether the workflow creates a snapshot or a partitioned table. The differences in these releases are entirely cosmetic. The function of the release class has been simplified to only concern itself with run-specific parameters/variables. The release class is only necessary because some aspects of the Workflows cannot be determined until runtime, so the release class should reflect only these aspects. The Workflow release classes thus tend to have little to no class functions and contain only filepaths and dates.

## DAG Creation
The DAG creation is now done from the observatory-platform's [load_workflows.py](https://github.com/The-Academic-Observatory/observatory-platform/blob/feature/workflow-config-file/observatory-platform/observatory/platform/dags/load_workflows.py). Since this loads the DAGs in a standard way, there is no room to customise individual DAG loads. Instead, a .yaml config file is used to describe each individual workflow (DAG). This makes the workflow loading much more transparent. DAG IDs are supplied explicitly in the config file and passed to the Workflow constructor. This is also for the purpose of transparency. Prior to this update, DAG IDs were constructed from the DAG prefix and the Organisation - which is now deprecated.

## Onix Workflow
The Onix Workflow is by far the most complex of the workflows. At present, we have only a handful of data sources. Since all data sources must be accounted for and aggregated in the Onix Workflow, an increase in our operations and data sources will quickly inflate the length and complexity of this workflow. The PR has made some attempts to mitigate the complexity and reduce volume of hard-coding. This is still rampant throughout the workflow and more effort is needed. Since this was not the primary purpose of this refactor, I have tried to limit the time I spent and have forgone any fundamental changes to the workflow. Now, it should be much easier to read and follow. This is also the case for the Onix Workflow Test, which had a large volume of Mock tests. I have elected to remove these as they appeared to emphasize testing of the implementation of the class functions, rather than their functionality.

## OaebuPartner
The oaebu_partners.py file has been moved from the workflow directory to its parent directory. The OaebuPartner class has been slightly changed and the file now contains a hard-coded list of partners. This list is used as a lookup - the partner names are supplied in the new config file and passed to the onix workflow upon DAG creation.

## SQL and Table ID
Some of the SQL (jinja templated) has been updated. This is to accommodate the new table format. The fully qualified table ID is passed now, rather than the separate project ID, dataset and table name. This makes both the SQL template and the python code much cleaner and easier to read. If we want access to any specific parts of the FQ table ID in the template, these can be supplied separately.

## Release Date Terminology
We have updated the terminology used for the release date. In most cases, we refer to either the _snapshot_ or _partition_ date. These can be thought of as a more specific case of a _release date_. In some cases, we may still refer to these instances as a _release date_. We should try to be as specific as possible, but for all intents and purposes, a _snapshot/parition date_ is a _release date_.


oaebu_workflows/workflows/tests/test_elastic_workflow.py
```diff
@@ -152,18 +150,27 @@ def test_load_elastic_mappings_oaebu(self):
...
+                Workflow(
+                    dag_id="elastic_import_test",
+                    name="Elastic Import Workflow",
+                    class_name="observatory.platform.workflows.elastic_import_workflow.ElasticImportWorkflow",
+                    cloud_workspace=CloudWorkspace(
+                        project_id=self.project_id,
+                        download_bucket="download-bucket",
+                        transform_bucket="transform-bucket",
+                        data_location=self.data_location,
+                    ),
```
jdddog, line 166: Can replace with `cloud_workspace=self.fake_cloud_workspace,`

oaebu_workflows/tests/test_seeding.py
```diff
@@ -16,7 +16,7 @@
 # Author: Tuan Chien
 
 
-from observatory.platform.utils.test_utils import ObservatoryTestCase, find_free_port
+from observatory.platform.observatory_environment import ObservatoryTestCase, find_free_port
```
jdddog, line 19: This file can be deleted

oaebu_workflows/workflows/fulcrum_telescope.py
```diff
@@ -16,86 +16,67 @@
...
-        source_format: str = SourceFormat.NEWLINE_DELIMITED_JSON,
-        airflow_conns: List = None,
-        catchup: bool = True,
-        download_file_name="fulcrum.jsonl.gz",
-        transform_file_name="fulcrum.jsonl.gz",
-        dataset_description: str = "IRUS Fulcrum Data Feed",
-        load_bigquery_table_kwargs=None,
+        api_dataset_id: str = "fulcrum",
+        observatory_api_conn_id: str = AirflowConns.OBSERVATORY_API,
+        oapen_irus_api_conn_id: str = "oapen_irus_uk_api",
```
jdddog, line 79: Some general feedback for the constructors to make them consistent with what I've done so far:
* I think we can just remove `load_bigquery_table_kwargs` and `source_format` (which you've done in this one) from all of the constructors and directly set the values in the respective `bq_load_table` functions. I think that it was only really necessary with the old design.
* The table description should be none.
* The docstrings should be updated.
* It would be good to move the Airflow specific parameters to the end, e.g. `start_date`, `schedule_interval`.
* We might need to make a helper function to assert that `dag_id` is not None or an empty string and that `cloud_workspace` is not None. cloud_workspace can be None. What do you think?

oaebu_workflows/workflows/fulcrum_telescope.py
```diff
@@ -177,131 +140,91 @@ def make_release(self, **kwargs) -> FulcrumRelease:
         """Create a FulcrumRelease instance"""
         # Release date should be the month before the execution date
         release_date = make_release_date(**kwargs).subtract(months=1)
```
jdddog, line 142: Why is the release date the month before the execution date? As the execution date is already the start of the data interval. 

I though that it was easier to change `make_snapshot_date` to just use the `data_interval_end`. The data interval uses a semi open interval [data_interval_start, data_interval_end) where the data_interval_end is not included.

oaebu_workflows/workflows/fulcrum_telescope.py
```diff
@@ -177,131 +140,91 @@ def make_release(self, **kwargs) -> FulcrumRelease:
...
     def download(self, release: FulcrumRelease, **kwargs):
         """Task to download the Fulcrum data for a release
 
         :param releases: the FulcrumRelease instance.
         """
-        requestor_id = BaseHook.get_connection(AirflowConns.OAPEN_IRUS_UK_API).login
-        download_month = release.release_date.format("YYYY-MM")
+        requestor_id = BaseHook.get_connection(self.oapen_irus_api_conn_id).login
+        download_month = release.partition_date.format("YYYY-MM")
         totals_data, country_data = download_fulcrum_month_data(download_month, requestor_id)
```
jdddog, line 152: I would probably change download_fulcrum_month_data so you pass it a pendulum date and it does the .format inside.

oaebu_workflows/workflows/fulcrum_telescope.py
```diff
@@ -177,131 +140,91 @@ def make_release(self, **kwargs) -> FulcrumRelease:
...
-            schema_file_path=schema_file_path,
-            project_id=self.project_id,
-            transform_bucket=self.transform_bucket,
-            transform_blob=transform_blob,
-            dataset_id=self.dataset_id,
-            data_location=self.data_location,
+        uri = gcs_blob_uri(self.cloud_workspace.transform_bucket, gcs_blob_name_from_path(release.transform_path))
+        table_id = bq_table_id(self.cloud_workspace.project_id, self.bq_dataset_id, self.bq_table_name)
+        schema_file_path = bq_find_schema(path=self.schema_folder, table_name=self.bq_table_name)
+        bq_load_table(
```
jdddog, line 203: You should also check if `bq_load_table` was successful or not with set_task_state(success, self.bq_load.__name__, release)

oaebu_workflows/workflows/fulcrum_telescope.py
```diff
@@ -120,49 +101,31 @@ def __init__(
...
+        self.bq_dataset_description = bq_dataset_description
+        self.bq_table_description = bq_table_description
+        self.bq_dataset_id = bq_dataset_id
+        self.bq_table_name = bq_table_name
+        self.schema_folder = schema_folder
+        self.load_bigquery_table_kwargs = load_bigquery_table_kwargs
+
+        self.api_dataset_id = api_dataset_id
+        self.oapen_irus_api_conn_id = oapen_irus_api_conn_id
+        self.observatory_api_conn_id = observatory_api_conn_id
```
jdddog, line 128: The cleanup task should come after add_new_dataset_releases

oaebu_workflows/workflows/tests/test_fulcrum_telescope.py
```diff
@@ -134,32 +134,18 @@ def setup_connections(self, env):
...
-            download_bucket="download_bucket",
-            transform_bucket="transform_bucket",
+        cloud_workspace = CloudWorkspace(
+            project_id=self.project_id,
+            download_bucket="download-bucket",
+            transform_bucket="transform-bucket",
             data_location=self.data_location,
+        )
+        dag = FulcrumTelescope(
+            dag_id="fulcrum_test", cloud_workspace=cloud_workspace, publishers=FAKE_PUBLISHERS
```
jdddog, line 148: You can replace `cloud_workspace=cloud_workspace` with `cloud_workspace=self.cloud_workspace` now for `test_dag_structure` and `test_dag_load`.

oaebu_workflows/workflows/tests/test_fulcrum_telescope.py
```diff
@@ -178,111 +164,121 @@ def test_dag_structure(self):
...
+                    kwargs=dict(publishers=[FAKE_PUBLISHERS]),
+                )
+            ]
+        )
         with env.create():
-            self.setup_connections(env)
-            self.setup_api()
-            dag_file = os.path.join(module_file_path("oaebu_workflows.dags"), "fulcrum_telescope.py")
-            self.assert_dag_load(f"fulcrum_university_of_michigan_press", dag_file)
+            self.assert_dag_load_from_config("fulcrum_test")
```
jdddog, line 184: If there are any required keyword arguments that are specified in kwargs, it would be best to check that an AssertionError is raised:

```python
        # Failure to load caused by missing kwargs
        env = ObservatoryEnvironment(
            workflows=[
                Workflow(
                    dag_id=self.dag_id,
                    name="Scopus Telescope Curtin University",
                    class_name="academic_observatory_workflows.workflows.scopus_telescope.ScopusTelescope",
                    cloud_workspace=self.fake_cloud_workspace,
                    kwargs=dict(),
                )
            ]
        )

        with env.create():
            with self.assertRaises(AssertionError) as cm:
                self.assert_dag_load_from_config(self.dag_id)
            msg = cm.exception.args[0]
            self.assertTrue("missing 2 required keyword-only arguments" in msg)
            self.assertTrue("institution_ids" in msg)
            self.assertTrue("scopus_conn_ids" in msg)
```

oaebu_workflows/workflows/tests/test_fulcrum_telescope.py
```diff
@@ -67,7 +68,6 @@ def __init__(self, *args, **kwargs):
         super(TestFulcrumTelescope, self).__init__(*args, **kwargs)
         self.project_id = os.getenv("TEST_GCP_PROJECT_ID")
         self.data_location = os.getenv("TEST_GCP_DATA_LOCATION")
-        self.data_location = "us"
 
         # API environment
         self.host = "localhost"
```
jdddog, line 73: You can remove all of the API environment stuff from the constructor as well as the setup_api and setup connections methods.

oaebu_workflows/workflows/tests/test_fulcrum_telescope.py
```diff
@@ -178,111 +164,121 @@ def test_dag_structure(self):
...
-            self.setup_connections(env)
-            self.setup_api()
-            dag_file = os.path.join(module_file_path("oaebu_workflows.dags"), "fulcrum_telescope.py")
-            self.assert_dag_load(f"fulcrum_university_of_michigan_press", dag_file)
+            self.assert_dag_load_from_config("fulcrum_test")
 
-    @patch("oaebu_workflows.workflows.fulcrum_telescope.ORG_PUBLISHER_MAPPINGS", FAKE_ORG_PUBLISHER_MAPPINGS)
     def test_telescope(self):
         """Test the Fulcrum telescope end to end."""
         # Setup Observatory environment
```
jdddog, line 188: Take a look at test_telescope to get an idea about what can be changed here: https://github.com/The-Academic-Observatory/academic-observatory-workflows/blob/feature/workflow-config-file/academic_observatory_workflows/workflows/tests/test_crossref_metadata_telescope.py#L105

oaebu_workflows/workflows/tests/test_fulcrum_telescope.py
```diff
@@ -178,111 +164,121 @@ def test_dag_structure(self):
...
+                project_id=env.project_id,
                 download_bucket=env.download_bucket,
                 transform_bucket=env.transform_bucket,
-                data_location=self.data_location,
-                dataset_id=dataset_id,
+                data_location=env.data_location,
+            )
+            telescope = FulcrumTelescope(
+                dag_id="fulcrum_test",
+                cloud_workspace=cloud_workspace,
```
jdddog, line 206: Can replace `cloud_workspace=cloud_workspace` with `env.cloud_workspace`

oaebu_workflows/workflows/tests/test_fulcrum_telescope.py
```diff
@@ -178,111 +164,121 @@ def test_dag_structure(self):
...
+                bq_dataset_id=env.add_dataset(),
             )
             dag = telescope.make_dag()
 
             # Add the fake requestor ID as a connection
             with env.create_dag_run(dag, execution_date):
                 # Test that all dependencies are specified: no error should be thrown
-                ti = env.run_task(telescope.check_dependencies.__name__)
-                self.assertEqual(ti.state, State.SUCCESS)
+                env.run_task(telescope.check_dependencies.__name__)
```
jdddog, line 215: You should always test that the task state is successful, as it could be in an invalid state, e.g. skipped, or failure.

```
ti = env.run_task(telescope.check_dependencies.__name__)
self.assertEqual(State.SUCCESS, ti.state)
```

oaebu_workflows/workflows/tests/test_fulcrum_telescope.py
```diff
@@ -178,111 +164,121 @@ def test_dag_structure(self):
...
+                self.assert_cleanup(release.download_folder, release.extract_folder, release.transform_folder)
 
                 # add_dataset_release_task
                 dataset_releases = get_dataset_releases(dataset_id=1)
                 self.assertEqual(len(dataset_releases), 0)
-                ti = env.run_task("add_new_dataset_releases")
+                env.run_task("add_new_dataset_releases")
                 self.assertEqual(ti.state, State.SUCCESS)
                 dataset_releases = get_dataset_releases(dataset_id=1)
                 self.assertEqual(len(dataset_releases), 1)
```
jdddog, line 284: To test dataset releases and cleanup, you can generally use this snippet, when there is only 1 release:
```python
# Test that DatasetRelease is added to database
dataset_releases = get_dataset_releases(dag_id=self.dag_id, dataset_id=workflow.api_dataset_id)
self.assertEqual(len(dataset_releases), 0)
ti = env.run_task(workflow.add_new_dataset_releases.__name__)
self.assertEqual(State.SUCCESS, ti.state)
dataset_releases = get_dataset_releases(dag_id=self.dag_id, dataset_id=workflow.api_dataset_id)
self.assertEqual(len(dataset_releases), 1)

# Test that all workflow data deleted
ti = env.run_task(workflow.cleanup.__name__)
self.assertEqual(State.SUCCESS, ti.state)
self.assert_cleanup(release.workflow_folder)
````

**Review by jdddog — CHANGES_REQUESTED** [review]
Hey Keegan, thanks for refactoring the code, it is looking very good. I've left some more general feedback on a couple of the workflows and tests.

File: oaebu_workflows/dags/onix_workflow.py [hunk 15]
```diff
@@ -21,7 +21,7 @@
 
 from airflow.exceptions import AirflowException
 from oaebu_workflows.identifiers import WorkflowTypes
-from oaebu_workflows.workflows.oaebu_partners import OaebuPartner
+from oaebu_workflows.oaebu_partners import OaebuPartner
 from oaebu_workflows.workflows.onix_workflow import OnixWorkflow
```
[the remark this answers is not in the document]
keegansmith21: OaebuPartners? This should still be required as the config file points to each partner when loading the onix workflow dag

oaebu_workflows/workflows/google_analytics_telescope.py
```diff
@@ -377,8 +432,8 @@ def get_dimension_data(
                 "pageSize": 10000,
                 "dateRanges": [
                     {
-                        "startDate": start_date.strftime("%Y-%m-%d"),
-                        "endDate": end_date.strftime("%Y-%m-%d"),
+                        "startDate": data_interval_start.strftime("%Y-%m-%d"),
+                        "endDate": data_interval_end.strftime("%Y-%m-%d"),
                     }
```
jdddog, line 437: Does the query with the dateRanges parameter include the endDate? [data_interval_start, data_interval_end) is a half open interval, so data_interval_start <= valid dates < data_interval_end.

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -258,15 +193,27 @@ def make_release(self, **kwargs) -> List[JstorRelease]:
...
     def check_dependencies(self, **kwargs) -> bool:
         """Check dependencies of DAG. Add to parent method to additionally check for a publisher id
+
         :return: True if dependencies are valid.
         """
         super().check_dependencies()
-
         if self.publisher_id is None:
             expected_extra = {"publisher_id": "jstor_publisher_id"}
             raise AirflowException(f"Publisher ID is not set in 'extra' of telescope, extra example: {expected_extra}")
```
jdddog, line 219: I don't think this is necessary anymore because publisher_id is specified in the constructor and the DAG won't be created if it isn't specified.

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -299,59 +241,185 @@ def download_reports(self, **kwargs) -> bool:
...
+        )
 
-    :return: headers dictionary
-    """
+        for release in releases:
+            for table_name, table_description, file_path in zip(
+                [self.bq_country_table_name, self.bq_institution_table_name],
+                [self.bq_country_table_description, self.bq_institution_table_description],
+                [release.transform_country_path, release.transform_institution_path],
+            ):
```
jdddog, line 332: Perhaps more straightforward to remove the zip and do this:
[(self.bq_country_table_name, self.bq_country_table_description, release.transform_country_path), (self.bq_institution_table_name, self.bq_institution_table_description, release.transform_institution_path)]

oaebu_workflows/workflows/onix_workflow.py
```diff
@@ -16,56 +16,52 @@
 # Author: Tuan Chien & Richard Hosking
 
 import os
-import shutil
 from datetime import timedelta
-from functools import partial, update_wrapper
-from pathlib import Path
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Union
```
jdddog, line 20: Nice work on simplifying the Onix Workflow!

jdddog (changes requested):
Hey Keegan, the changes look good, I've just left a few comments. It would also be good to set the partition_date field in the API for the oaebu workflows that use partioned tables.

oaebu_workflows/workflows/onix_workflow.py
keegansmith21: Thanks! It was definitely the most challenging part of the refactor. I think it can still be better, but it's much cleaner now :smile:

oaebu_workflows/workflows/jstor_telescope.py
keegansmith21: Oh yeah... I wonder why I'd never done it this way before. Thanks for the suggestion!

jdddog (approved):

PR #143, opened 2023-03-31
BAD-507: Add personname field to Thoth feed

keegansmith21:
The Thoth ONIX feeds do not have the personname field populated by default. This PR uses the same function created for OAPEN (#135) to create the field from two other fields (should they exist)


jdddog (approved):
Thanks Keegan, looks good.

PR #144, opened 2023-05-12
Feature/crossref academic observatory

keegansmith21:
Since removing the dependency on the [Academic Observatory Workflows](https://github.com/The-Academic-Observatory/academic-observatory-workflows) in #105, we have run into several issues with the collection of crossref data. The crux of the issue is the volume of requests we need to make to the APIs (metadata and events). This PR will instead reintroduce the dependency on the AO workflows and query the Crossref Metadata master table. Noteably, this is a different solution as was used previously to #105. However, I believe that using the metadata table over the book table is preferable for transparency and simplicity. The caveat is that the metadata table is significantly larger and querying it once a week for each publisher will eventually become a noticeable cost. There are ways to mitigate this cost, but for the immediate future, this is a simple solution. 

Querying the metadata table is done with a jinnja2-templated SQL statement, where each ISBN of interest is inserted into a temporary table which is subsequently joined against the metadata table's ISBN field. I have also included a similar file for the crossref events table in anticipation of its future use.

I have had to commit a sin in order to get the tests to pass. In the upcoming refactor #142 this will be removed in favour of a more complete table-checking method, so this is only temporary.


jdddog (approved):
Thanks Keegan, it looks good.

I would just create the Crossref Metadata sub-table by saving the output of the query directly to a BigQuery table, but that can be done later.

PR #146, opened 2023-05-29
Remove ISBN filter on OA books

keegansmith21:
We are transitioning to the option of processing all book titles - rather than just digital open access. For publishers that want only OA titles, the supplied ONIX feed remains the source of truth.


jdddog (approved):
Thanks Keegan, looks good!

PR #148, opened 2023-06-07
Google books sftp update

keegansmith21:
For the same reasons as #147, the sftp root needs to be changed


jdddog (approved):
Thanks Keegan, looks good.

PR #149, opened 2023-06-08
Changed partition write disposition to append

keegansmith21:
The write disposition should be set to append for all tables. The default behaviour is to overwrite, which would be destructive for our data.


jdddog (approved):
Thanks Keegan, looks good.

PR #152, opened 2023-06-15
Fixed file saving for empty contents

keegansmith21:
Bugfix for when a GB report has no contents. The issue was that the file would not be written if the report is empty - then the upload task would raise an error since it can't find the file.


jdddog (approved):
Thanks Keegan, looks good

PR #153, opened 2023-06-15
Fix GA schema find

keegansmith21:
The google analytics find_schema needs to search foir the organisation name prefix. Otherwise it'll find the base schema which leads to bigquery loading errors


jdddog (approved):
Thanks Keegan, looks good

PR #154, opened 2023-06-15
Fixed the file matching regex

keegansmith21:
The date matching regex was being used for the file match on the server onix files. This would not match on the file as it's meant to be used as a search. I altered the regex match to just match on the entire files


jdddog (approved):
Thanks Keegan, looks good.

PR #157, opened 2023-06-22
fixed jstor write disposition

keegansmith21:
JSTOR was overwriting its main table instead of appending to it


jdddog (approved):
Thanks Keegan, looks good

PR #161, opened 2023-06-26
Removed additional table load in test

keegansmith21:
The PR from the observatory platform (The-Academic-Observatory/observatory-platform#623) has revealed a small bug in one of the oapen metadata telescope tests. This PR fixes it.


jdddog (approved):
Thanks Keegan, looks good.

PR #162, opened 2023-07-13
IRUS schedule change

keegansmith21:
Changed the IRUS run dates to be the 4th of each month as this is typically when they're available by.


jdddog (approved):
Thanks Keegan, looks good.

PR #160, opened 2023-06-23
oapen irus uk rename to irus oapen

keegansmith21:
Name changes as requested in #158 

## Notes for post-merge
- Change dataset names in bigquery
- Change table names in bigquery
- Change dag_ids of historic releases in API
- Change name of irus airflow connections

PR #165, opened 2023-07-31
Rename empty IRUS tables

keegansmith21:
Quick fix for the IRUS empty tables. Missed in the original PR #160 because the tests don't cover empty data.


jdddog (approved):
Thanks Keegan, looks good!

PR #166, opened 2023-08-01
Updated oapen metadata to work with additional OAPEN data sources

keegansmith21:
# OAPEN Metadata Telescope Update
This PR updates the OAPEN Metadata telescope with one additional line of functionality - the ability to change the source URI of the metadata. This has become necessary as the BTOB collection (and upcoming SHMP collection) will host their metadata on the OAPEN platform. This change in itself was not substantial, the majority of the altered code has been a refactor of how the xmls are handled and how invalid products are removed from them. I have also updated the tests to be more verbose.


jdddog (approved):
Thanks Keegan, looks good.

PR #169, opened 2023-08-04
Fix/jstor heading update

keegansmith21:
JSTOR has slightly changed their formatting, but new reports can still be processed in the same manner. Just needed to change the nature of the check for old vs. new formatted reports.


jdddog (approved):
Looks good Keegan

PR #164, opened 2023-07-19
Update of UCL discovery telescope

keegansmith21:
# UCL Discovery

## PR Necessity
This PR is a major update of the UCL Discovery telescope. This update has been necessitated by the manual intervention that has been required to harvest the entirety of the discovery repository's data. Due to the incompleteness of discovery's metadata, as harvested from their API, this intervention has been required. Since there is no way to retrieve the entirety of the metadata, we instead opt to switch to a semi-manual approach.

## Google Sheet
The metadata of each title has been, and continues to be provided to us via a contact at UCL through email interface. This is how we have maintained a complete metadata collection. This relies heavily on manual interference. As such, we have moved our current metadata collection to a Google Sheet. This presents the opportunity to both structure the data, and restrict its access to only those who would require it. Since Google Sheets have a simple interface, no substantial technical knowledge is required to enter the data (which should be done on a monthly basis). The is contrary to alternative solutions such as using the already-existing SFTP server. The telescope utilises the Python library for the Google Sheets API to extract the data. This relies on the sheet being formatted specifically and any major change to the structure of the sheet will likely require an update to the telescope.

## Sheet Metadata
The metadata that is stored on the google sheet is minimal. Only a few fields are required for the telescope to function. The most imperative is the relationship between the UCL proprietary identifier, _eprintID_, and the universal book identifier, _ISBN13_. We require the eprint ID to query the UCL discovery usage API and collect the monthly usage data. We then map this to the correct ISBN13 for future data processing by the onix_workflow.

## Caveats
It should be noted that the telescope's logical workflow does present a possible issue. If the stored in the sheet is incomplete at the time it's run, then the incomplete titles will be ignored and the telescope will run without collecting their data. I have elected to forgo throwing an error as there are some titles that we wish to store on the sheet that are not mapped to an ISBN, or that have not yet been published and _should_ be ignored. In the case that the metadata is not updated and the telescope does run without all titles, we will need to manually remove the partition from the table and rerun the DAG.

PR #170, opened 2023-08-10
Fixed empty table creation in SQL

keegansmith21:
When updating #164 I didn't account for the empty tables - which are apparently incorrectly formatted. I also made missing country fields null instead of "N/A" as I think that might flow through as a literal country called "N/A".


jdddog (approved):
Looks good Keegan!

PR #171, opened 2023-08-10
Remove Elastic mappings

jdddog:
Remove Elasticsearch mappings and functions which are no longer used.


keegansmith21 (approved):
This looks so clean now! Thanks Jamie!

PR #173, opened 2023-08-18
Moved crossref dataset creation to metadata table task

keegansmith21:
The crossref dataset creation was under the second-run crossref task - so the first one would run into an error if it doesn't already exist


jdddog (approved):
Thanks Keegan

PR #185, opened 2023-09-13
Include published date field

keegansmith21:
Added the published_date field to the book_products and export_book_product_list tables. For the latter, this takes the place of the time_field that is no longer in use.


kathrynnapier (approved):
Thanks Keegan, looks good! 

Would it make sense to make the descriptions consistent between oaebu_workflows/database/schema/book_product.json and oaebu_workflows/database/schema/oaebu_publisher_book_product_list.json ?

e.g. "The date the book was published" vs "The date of publication of the book"

This could be addressed later as a future improvement if there are more inconsistencies between the two schemas.

kathrynnapier:
Noting that @kathrynnapier and @bechandcock will do a sprint through current BAD dashboards to remove any (erroneous) inclusions of time_field in visualisations/filters (not a required field, but a hangover from previous kibana visualisations)

keegansmith21:
> Thanks Keegan, looks good!
> 
> Would it make sense to make the descriptions consistent between oaebu_workflows/database/schema/book_product.json and oaebu_workflows/database/schema/oaebu_publisher_book_product_list.json ?
> 
> e.g. "The date the book was published" vs "The date of publication of the book"
> 
> This could be addressed later as a future improvement if there are more inconsistencies between the two schemas.

Thanks Kathryn! There are probably quite a lot of these little inconsistencies between the schemas. But I do agree that it's better to try to keep them all the same. I'll change it here and perhaps we can address this issue with the upcoming refactor.

alexmassen-hane (approved):
Looks good Keegan.

PR #186, opened 2023-09-14
Added InvertedPersonName field

keegansmith21:
Added the InvertedPersonName fields to the book_product table. This flows through to the author_metrics export table. Much like the PersonName field, this field is created for the oapen_metadata and thoth feeds where possible by concatenating the first and last names.

Also added the author information to the book_product_list export table.

### DIFF
--- a/oaebu_workflows/database/sql/export_book_author_metrics.sql.jinja2
+++ b/oaebu_workflows/database/sql/export_book_author_metrics.sql.jinja2
@@ -26,6 +26,7 @@ CREATE TEMP FUNCTION group_counts(counts ARRAY<STRUCT<name STRING, value INT64>>
 
 SELECT
     author.PersonName,
+    MAX(author.PersonNameInverted) as PersonNameInverted,
     MAX(author.ORCID) as orcid,
     count(ISBN13) as unique_books,
     month.month,
--- a/oaebu_workflows/database/schema/book_product.json
+++ b/oaebu_workflows/database/schema/book_product.json
@@ -73,7 +73,13 @@
             "mode": "NULLABLE",
             "name": "PersonName",
             "type": "STRING",
-            "description": "The Authors Full Name"
+            "description": "The Author's Full Name in the format '[first name] [last name]'"
+          },
+          {
+            "mode": "NULLABLE",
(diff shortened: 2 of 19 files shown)


oaebu_workflows/database/sql/export_book_author_metrics.sql.jinja2
```diff
@@ -67,6 +68,6 @@ SELECT
     ) as irus_fulcrum,
 FROM `{{ book_product_table_id }}`, UNNEST(onix.authors) as author, UNNEST(months) as month
 WHERE month.irus_fulcrum IS NOT NULL OR month.irus_oapen IS NOT NULL OR month.google_books_traffic IS NOT NULL OR month.google_books_sales IS NOT NULL
-GROUP BY author.PersonName, month
+GROUP BY author.PersonName, author.PersonNameInverted, month
 ORDER BY author.PersonName ASC, month DESC
```
kathrynnapier, line 72: Hi @keegansmith21 - I haven't looked into the population of the PersonNameInverted field, but in the instance there are any data quality issues (e.g. more than one PersonNameInverted mapping back to a PersonName, empty PersonNameInverted field), would MAX(author.PersonNameInverted) be a safer GROUP BY?

kathrynnapier (approved):
Thanks @keegansmith21 ! Just had a query about the addition of PersonNameInverted in the GROUP BY when aggregating at the author level, in case there are issues with data quality and the potential for a one to many relationship between PersonName and PersonNameInverted duplicating the aggregation. Your suggestion of MAX(PersonNameInverted) should work well I think.

oaebu_workflows/database/sql/export_book_author_metrics.sql.jinja2
keegansmith21: HI @kathrynnapier - Thanks for pointing that out. You're right - if the personname field is populated but the inverted personname is not, this will duplicate the author. I have altered the aggregation method of the inverted field in the most recent commit. Would you be able to give it a once-over? :smile:

kathrynnapier (commented):
Thanks @keegansmith21 - looks good to me.

PR #188, opened 2023-09-19
templated ga3 views type

keegansmith21:
We changed the field to page_views from unique_views in #95. We now need access to either the unique OR page views depending on the publisher. To accomplish this, I altered the book_product sql to template out the google_analytics views field. This requires the views_field to be supplied when rendering - it defaults to page_views in the ONIX_WF telescope.

Note - we'll have to add the ga3_views_field to the config file ONIX WFs that contain google_analytics as data partner.


alexmassen-hane (approved):
Thanks Keegan, looks good. :+1:

PR #189, opened 2023-09-27
Fix published_date for YYYY entries

keegansmith21:
Some metadata feeds have the publishing date in YYYY format - which is perfectly acceptable by ONIX standards. The book_product query breaks when encountering something like this though. This PR changes the query to ignore the published_date field if the ONIX PublishingDate field is not in YYYYMMDD format (ie. 8 characters long).


jdddog (approved):

PR #190, opened 2023-09-29
Fix issue with OAPEN metadata filter

keegansmith21:
This PR addresses a bug with the OAPEN metadata filter (formerly called the parser). The bug would skip over lists, which was a big issue as the main field we want to filter on is the list of products. The tests did not attempt to filter any lists, so this has also been amended. I've also updated the naming convention as I think 'filter' is more apporpriate and sets it apart from the actual parsing step.


jdddog (approved):
Thanks Keegan, looks good.

PR #194, opened 2023-10-13
Related Products Elevation & Normalisation

keegansmith21:
This PR was necessitated by two related quirks of the OAPEN metadata feed:

1. A product's main identifier is often that of the print version of the product instead of the open access version. Instead, the open access version may be stored as a Related Product of the main product.
2. The Related Products are stored in a malformed manner according to ONIX 3.0 specifications. This was resulting in all related products but 1 being dropped during the parsing stage of transformation.

To fix these two abnormalities I have implemented the following solutions:
## 1. Related Product Elevation
For every _RelatedProduct_ in the metadata feed, a new product is created that has this related product as its main identifier, and swaps out the RP identifier for that of the main product. This step is technically introducing fake metadata, so we only want to do it when necessary and definitely should not be applied to all publishers.

## 2. Related Product Normalisation
ONIX 3.0 specifies that each _RelatedProduct_ should have only one _ProductIdentifier_. There is an exception to be made when there are identical _ProductIdentifier_ elements with different _ProductIDType_ values. 
This step will pull out unnecessary _ProductIdentifier_ elements into their own _RelatedProduct_ elements to conform with specification.
This step should also only be necessary for select sources and not implemented into all ONIX transformations.

## ONIX Transform Consolidation
With the addition of several new functions and steps in the oapen_metadata_telescope's transform step, the state of the transform code had become messy. I have renamed the onix.py to onix_utils.py and its purpose is to store functions relating to the handling of metadata for the telescopes. I have also moved many of the functions that were in oapen_metadata_telescope.py to this file. 

To deal with the growing transform steps for oapen metadata, I have created the OnixTransformer class (in onix_utils). This class has only a single function made for use - the transform function. When called, this will execute the transform stage based on the configuration of the class instantiated during object initialisation.
This class can also be used for the other ONIX transform steps (onix_telescope.py and thoth.py). I have not implemented this change as the scope of this PR is already quite extensive, but it would be good to make use of it in future.


alexmassen-hane (approved):
Thanks Keegan, looks good.

oaebu_workflows/onix_utils.py
```diff
@@ -0,0 +1,714 @@
...
+        return self._current_md_path
+
+    def _intermediate_file_path(self, file_name):
+        dir_ = self.output_dir if self.keep_intermediate else self._work_dir
+        return os.path.join(dir_, file_name)
+
+    def _save_metadata(self, metadata: List[dict], file_path: str):
+        save_path = os.path.join(self._work_dir, file_path)
+        format = re.search(r"\.(.*)$", file_path).group(1)
+        with open(save_path, "w") as f:
```
jdddog, line 177: You can remove the outer `with open(save_path, "w") as f:` and put it under `if format == "xml":`

oaebu_workflows/onix_utils.py
```diff
@@ -0,0 +1,714 @@
...
+            if format == "xml":
+                xmltodict.unparse(metadata, output=f, pretty=True)
```
jdddog, line 179: I get the warning:

```python
Unexpected type(s): (list[dict], TextIO, bool) Possible type(s): (Mapping[str, Any], SupportsWrite[bytes] | SupportsWrite[str], Any) (Mapping[str, Any], None, Any) 
```

oaebu_workflows/onix_utils.py
```diff
@@ -0,0 +1,714 @@
...
+        elif format == "jsonl" or format == "jsonl.gz":
+            metadata = load_jsonl(file_path)
+        else:
+            raise ValueError(f"Unsupported format: {format}")
+        return metadata
+
+    def _filter_products(self):
+        with open(self.filter_schema, "r") as f:
+            schema = json.load(f)
+        filtered_metadata = filter_through_schema(self.current_metadata, schema)
```
jdddog, line 211: The types seem incorrect here: `Expected type 'dict', got 'list[dict]' instead`

oaebu_workflows/onix_utils.py
```diff
@@ -0,0 +1,714 @@
...
+
+    def _remove_errors(self):
+        invalid_products_path = os.path.join(self.output_dir, self.invalid_products_name)
+        save_path = self._intermediate_file_path("errors_removed.xml")
+        remove_invalid_products(self._current_md_path, save_path, invalid_products_path)
+        self._current_md_path = save_path
+
+    def _normalise_related_products(self):
+        metadata = deepcopy(self.current_metadata)
+        fixed_products = normalise_related_products(metadata["ONIXMessage"]["Product"])
```
jdddog, line 222: I get the error: 

```python
Unexpected type(s): (str) Possible type(s): (SupportsIndex) (slice) 
```

oaebu_workflows/onix_utils.py
```diff
@@ -0,0 +1,714 @@
...
+                if rp_id == product_isbn and rp_id_type == "15":  # Ignore parent ISBN in related products
+                    continue
+                if rp_id not in seen_ids.keys():
+                    seen_ids[rp_id] = set()
+                seen_ids[rp_id].add(rp_id_type)
+
+        # Rebuild the related products
+        new_related_products = []
+        for id, id_types in seen_ids.items():
+            related_product = {"ProductRelationCode": relation_code, "ProductIdentifier": []}
```
jdddog, line 472: Is `relation_code` here meant to come from the item in `related_products`?

jdddog (changes requested):
Hey Keegan, I think it is largely good to go, just some small comments.

oaebu_workflows/onix_utils.py
keegansmith21: Ah yes good catch. In practice this makes little difference as OAPEN always gives a relation code of "06". But I've amended it with the latest commit.

jdddog (approved):
Thanks Keegan, thanks changes look good.

PR #196, opened 2023-10-16
BAD-646 updates to formatting after review from Laura. Changes made t…

kathrynnapier:
Formatting changes to Dashboard docs requested after review by Laura. Formatting changes also applied to oaebu workflows where applicable for consistency

Updates to rst files to remove <no title> bug in prior navigation paths.

### DIFF
--- a/docs/bad_project/overview/dashboard_data.md
+++ b/docs/bad_project/overview/dashboard_data.md
@@ -1,7 +1,10 @@
-Dashboard Data Sources
+Dashboard data sources
 ==========================
 
-The data sources currently visualised in the Dashboard are detailed in the table below. The standard data sources and variables used are included, other data sources and variables may be supported as an extra add-on service.
+## What are the Dashboard’s data sources?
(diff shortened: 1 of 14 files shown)


docs/bad_project/overview/dashboard_data.md
```diff
@@ -32,10 +35,10 @@ The public access data sources are those where data is made publicly available b
 ### Crossref Event Data
```
keegansmith21, line 35: should this be "Crossref event data"?

kathrynnapier: Thanks @keegansmith21 - Laura requested de capitalisation of common nouns, e.g. OAPEN metadata instead of OAPEN Metadata. If the data is named with initial letters capitalised, like Crossref Event Data, Google Books, then that should remain. I've found conflicting reference to Crossref Metadata vs Crossref metadata though. What do you think?

keegansmith21: Hm I would have though either both or neither of the Crossref sources to be capitalised. If Crossref itself isn't consistent then we should just choose whatever we like. I would have thought it would be "Crossref Metadata" and "Crossref Event data", but I'm not an expert :sweat_smile:

keegansmith21 (approved):
Thanks Kathryn!

PR #197, opened 2023-10-19
Add collections to JSTOR Telescope

keegansmith21:
JSTOR has added support for collections of works, as opposed to their normal functioning regime which involves grouping works by publisher.

From a technical perspective, Collection reports are a little different to publisher reports, but they are not different enough (and the metrics they provide are 1:1 comparable) to warrant a new telescope. Instead, I implemented an option for the current JSTOR telecope that makes room for processing a collection. The collection and publisher-unique functionality is contained by their own classes, which each inherit from the JstorAPI class.

### Data Partners
In the config, the following kwargs should be changed for JSTOR DAGs:
```diff
- publisher_id
+ entity_id: 
+ entity_type:
+ country_partner: 
+ institution_partner:
```
Where:
- **entity_id** is the name of the publisher or collection and is always required (previously called *publisher_id*)
- **entity_type** is one of either *publisher* or *collection* (defaults to publisher)
- **country_partner** is the name of the country partner (defaults to jstor_country)
- **institution_partner** is the name of the country partner (defaults to jstor_institution)

The *jstor_country_collection* and *jstor_institution_collection* data partners were added for the sole purpose of directing the telescope to the correct schema for table upload. The schemas are slightly different, so this was necessary. It is important to note, however, that the ONIX WF should use the original (jstor_country, jstor_institution) data partners **regardless**. This is because the two are treated as the same in the ONIX WF - the difference in fields is not important as it only processes ISBN usage.

### Regex:
The regex that matches on report name is provisional. I only have one example email to base this off so it may change as needed when onboarding future collections.


oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -573,84 +452,355 @@ def get_label_id(service: Resource, label_name: str) -> str:
...
-        original_email = f"https://mail.google.com/mail/u/0/#all/{message_id}"
-        if extension == "tsv":
-            # add report info
-            logging.info(
-                f"Adding report. Report type: {report_type}, url: {download_url}, " f"original email: {original_email}."
+            print(results[0].keys())
+            save_jsonl_gz(transform_file, results)
+
+
+class Publisher:
```
jdddog, line 584: The Publisher and Collections classes are an example where you could use inheritance to make the code easier to manage.

You can have a top level class called `JstorAPI`, add the common functionality to it and also add interfaces for the methods that need to be implemented by `JstorCollectionsAPI` and `JstorPublishersAPI`. See at the bottom for an example of the base class.

Then instead of having if statements like this throughout the workflow, you can just have one place where you decide what class will be used (e.g. in the constructor or make_release) and then refer to the same object and and call the same methods throughout the code, without the if statements.

```python
        if self.collection:
            available_reports = Collection.list_reports(service, self.publisher_id)
        else:
            available_reports = Publisher.list_reports(service, self.publisher_id)
```


```python
class JstorAPI(ABC):
    def __init__(self, service: Any):
        self.service = service

    @retry(stop=stop_after_attempt(JstorTelescope.MAX_ATTEMPTS), reraise=True, wait=JstorTelescope.WAIT_FN)
    def get_header_info(self, url: str) -> tuple[Any, Any]:
        """Get header info from url and parse for filename and extension of file.

        :param url: Download url
        :return: Filename and file extension
        """
        logging.info(
            f"Getting HEAD of report: {url}, "
            f'attempt: {Publisher.get_header_info.retry.statistics["attempt_number"]}, '
            f'idle for: {Publisher.get_header_info.retry.statistics["idle_for"]}'
        )
        headers = {"User-Agent": get_user_agent(package_name="oaebu_workflows")}
        response = requests.head(url, allow_redirects=True, headers=headers)
        if response.status_code != 200:
            raise AirflowException(
                f"Could not get HEAD of report download url, reason: {response.reason}, "
                f"status_code: {response.status_code}"
            )
        filename, extension = response.headers["Content-Disposition"].split("=")[1].split(".")
        return filename, extension

    def get_release_date_deprecated(self, report_path: str) -> tuple[DateTime, DateTime]:
        """This function is deprecated, because the headers for the reports have changed since 2021-10-01.
        It might still be used for reports that were created before this date and have not been processed yet.
        Get the release date from the "Usage Month" column in the first row of the report.
        Also checks if the reports contains data from the same month only.

        :param report_path: The path to the JSTOR report
        :return: The start and end dates
        """
        # Load report data into list of dicts
        with open(report_path) as tsv_file:
            csv_list = list(csv.DictReader(tsv_file, delimiter="\t"))

        # get the first and last usage month
        first_usage_month = csv_list[0]["Usage Month"]
        last_usage_month = csv_list[-1]["Usage Month"]

        # check that month in first and last row are the same
        if first_usage_month != last_usage_month:
            raise AirflowException(
                f"Report contains data from more than 1 month, start month: {first_usage_month}, "
                f"end month: {last_usage_month}"
            )

        # get the release date from the last usage month
        start_date = pendulum.from_format(first_usage_month, "YYYY-MM").start_of("month").start_of("day")
        end_date = pendulum.from_format(last_usage_month, "YYYY-MM").end_of("month").start_of("day")

        return start_date, end_date

    @abstractmethod
    def list_reports(self, publisher_id: str) -> List[Dict]:
        pass

    @abstractmethod
    def get_release_date(self, report_path: str) -> tuple[DateTime, DateTime]:
        pass

    @abstractmethod
    def transform_report(self, *args, **kwargs) -> None:
        pass

```

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -573,84 +452,355 @@ def get_label_id(service: Resource, label_name: str) -> str:
...
-            available_reports.append({"type": report_type, "url": download_url, "id": message_id})
-        else:
-            logging.warning(
-                f'Excluding file "{filename}.{extension}", as it does not have ".tsv" extension. '
-                f"Original email: {original_email}"
+        filename, extension = response.headers["Content-Disposition"].split("=")[1].split(".")
+        return filename, extension
+
+    @staticmethod
+    def get_release_date(report_path: str) -> pendulum.DateTime:
```
jdddog, line 699: The return types for many of these methods should be `tuple[pendulum.DateTime, pendulum.DateTime]`

jdddog (changes requested):
Thanks for the PR Keegan, it looks good. You could consolidate some of the functionality for the Publisher and Collections classes by using inheritance.

oaebu_workflows/workflows/jstor_telescope.py
keegansmith21: Thanks @jdddog , I've proceeded with this implementation. Please take a look and let me know what you think. 

I've also gone ahead and renamed a few things to be more accurate (such as *publisher_id* -> *entity_id*). The PR description has been updated to reflect these changes.

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -149,16 +153,26 @@ def __init__(
...
         self.observatory_api_conn_id = observatory_api_conn_id
 
+        # Create the Jstor API Instance
+        service = create_gmail_service()
+        if entity_type == "publisher":
+            self.jstor_api = JstorPublishersAPI(service, self.entity_id)
+        elif entity_type == "collection":
+            self.jstor_api = JstorCollectionsAPI(service, self.entity_id)
+        else:
+            raise AirflowException(f"Entity type must be 'publisher' or 'collection', got {entity_type}.")
```
jdddog, line 174: I would put this in a function and call it only when you need it from within the tasks, otherwise it will break the loading of all of the DAGs if someone doesn't specify the Airflow connection key.

```python
def make_jstor_api(entity_type: str, entity_id: str):
    # Create the Jstor API Instance
    service = create_gmail_service()
    if entity_type == "publisher":
        jstor_api = JstorPublishersAPI(service, entity_id)
    elif entity_type == "collection":
        jstor_api = JstorCollectionsAPI(service, entity_id)
    else:
        raise AirflowException(f"Entity type must be 'publisher' or 'collection', got {entity_type}.")

    return jstor_api
```

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -149,16 +153,26 @@ def __init__(
...
+
         check_workflow_inputs(self)
 
         self.add_setup_task(self.check_dependencies)
```
jdddog, line 178: Is the re-implementation of `check_dependencies` necessary?

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -216,8 +230,7 @@ def list_reports(self, **kwargs) -> bool:
 
         :return: Whether to continue the DAG
         """
-        service = create_gmail_service()
-        available_reports = list_reports(service, self.publisher_id)
+        available_reports = self.jstor_api.list_reports()
```
jdddog, line 233: An example of how to create the API instance.

```suggestion
        api = make_jstor_api(self.entity_type, self.entity_id)
        available_reports = api.list_reports()
```

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -289,7 +297,7 @@ def upload_downloaded(self, releases: List[JstorRelease], **kwargs) -> None:
     def transform(self, releases: List[JstorRelease], **kwargs):
         """Task to transform the Jstor releases for a given month."""
         for release in releases:
-            jstor_transform(
+            self.jstor_api.transform_reports(
                 download_country=release.download_country_path,
                 download_institution=release.download_institution_path,
                 transfrom_country=release.transform_country_path,
```
jdddog, line 303:
```suggestion
                transform_country=release.transform_country_path,
```

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -353,304 +361,488 @@ def add_new_dataset_releases(self, releases: List[JstorRelease], **kwargs) -> No
...
+    @abstractmethod
+    def get_release_date(self, report_path: str) -> tuple[pendulum.DateTime, pendulum.DateTime]:
+        pass
+
+    @abstractmethod
+    def download_report(self, report: dict, download_path: str) -> None:
+        pass
+
+    @abstractmethod
+    def transform_reports(self, *args, **kwargs) -> None:
```
jdddog, line 447:
```suggestion
    def transform_reports(
        self,
        download_country: str,
        download_institution: str,
        transform_country: str,
        transform_institution: str,
        partition_date: pendulum.DateTime,
    ) -> None:
        pass
```

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -353,304 +361,488 @@ def add_new_dataset_releases(self, releases: List[JstorRelease], **kwargs) -> No
...
+        pass
+
+    @abstractmethod
+    def add_labels(self, report_info: List[dict]) -> bool:
```
jdddog, line 451:
```suggestion
    def add_labels(self, reports: List[dict]) -> bool:
```

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -353,304 +361,488 @@ def add_new_dataset_releases(self, releases: List[JstorRelease], **kwargs) -> No
...
+                )
 
+        return available_reports
 
-def get_release_date_deprecated(report_path: str) -> pendulum.DateTime:
-    """This function is deprecated, because the headers for the reports have changed since 2021-10-01.
-    It might still be used for reports that were created before this date and have not been processed yet.
-    Get the release date from the "Usage Month" column in the first row of the report.
-    Also checks if the reports contains data from the same month only.
+    def download_report(self, report: str, download_path: str) -> None:
```
jdddog, line 527:
```suggestion
    def download_report(self, report: dict, download_path: str) -> None:
```

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -353,304 +361,488 @@ def add_new_dataset_releases(self, releases: List[JstorRelease], **kwargs) -> No
...
-            "name": label_name,
-            "messageListVisibility": "show",
-            "labelListVisibility": "labelShow",
-            "type": "user",
+    def transform_reports(
+        self,
+        download_country: List[dict],
+        download_institution: List[dict],
+        transfrom_country: str,
+        transform_institution: str,
```
jdddog, line 640:
```suggestion
        download_country: str,
        download_institution: str,
        transform_country: str,
        transform_institution: str,
```

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -353,304 +361,488 @@ def add_new_dataset_releases(self, releases: List[JstorRelease], **kwargs) -> No
...
+        """Transform a Jstor release into json lines format and gzip the result._summary_
+
+        :param download_country: The path to the country download report
+        :param download_institution: The path to the institution download report
+        :param transfrom_country: The path to write the transformed country file to
+        :param transform_institution: The path to write the transformed institution file to
+        :param partition_date: The partition/release date of this report
+        """
+        for download_file, transform_file in (
+            [download_country, transfrom_country],
```
jdddog, line 652:
```suggestion
            [download_country, transform_country],
```

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -353,304 +361,488 @@ def add_new_dataset_releases(self, releases: List[JstorRelease], **kwargs) -> No
...
+        download_institution: List[dict],
+        transfrom_country: str,
+        transform_institution: str,
+        partition_date: pendulum.DateTime,
+    ) -> None:
+        """Transform a Jstor release into json lines format and gzip the result._summary_
+
+        :param download_country: The path to the country download report
+        :param download_institution: The path to the institution download report
+        :param transfrom_country: The path to write the transformed country file to
```
jdddog, line 647:
```suggestion
        :param transform_country: The path to write the transformed country file to
```

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -353,304 +361,488 @@ def add_new_dataset_releases(self, releases: List[JstorRelease], **kwargs) -> No
...
+                if "Total_Item_Requests" not in transformed_row:  # Institution report names it "Reporting_Period_Total"
+                    transformed_row["Total_Item_Requests"] = transformed_row.pop("Reporting_Period_Total")
+                results.append(transformed_row)
+
+            results = add_partition_date(
+                results, partition_date, TimePartitioningType.MONTH, partition_field="release_date"
+            )
+            save_jsonl_gz(transform_file, results)
+
+    def add_labels(self, reports: List[str]) -> bool:
```
jdddog, line 682:
```suggestion
    def add_labels(self, reports: List[dict]) -> bool:
```

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -353,304 +361,488 @@ def add_new_dataset_releases(self, releases: List[JstorRelease], **kwargs) -> No
...
+        month = pendulum.from_format(dates[0], "MMMM YYYY")  # eg. September 2023
+        start_date = month.start_of("month").start_of("day")
+        end_date = month.end_of("month").start_of("day")
+        return start_date, end_date
+
+    def transform_reports(
+        self,
+        download_country: List[dict],
+        download_institution: List[dict],
+        transfrom_country: str,
```
jdddog, line 792:
```suggestion
        download_country: str,
        download_institution: str,
        transform_country: str,
```

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -353,304 +361,488 @@ def add_new_dataset_releases(self, releases: List[JstorRelease], **kwargs) -> No
...
+        """Transform a Jstor release into json lines format and gzip the result._summary_
+
+        :param download_country: The path to the country download report
+        :param download_institution: The path to the institution download report
+        :param transfrom_country: The path to write the transformed country file to
+        :param transform_institution: The path to write the transformed institution file to
+        :param partition_date: The partition/release date of this report
+        """
+        for entity, download_file, transform_file in (
+            ["Country_Name", download_country, transfrom_country],
```
jdddog, line 805:
```suggestion
            ["Country_Name", download_country, transform_country],
```

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -353,304 +361,488 @@ def add_new_dataset_releases(self, releases: List[JstorRelease], **kwargs) -> No
...
+        download_institution: List[dict],
+        transfrom_country: str,
+        transform_institution: str,
+        partition_date: pendulum.DateTime,
+    ) -> None:
+        """Transform a Jstor release into json lines format and gzip the result._summary_
+
+        :param download_country: The path to the country download report
+        :param download_institution: The path to the institution download report
+        :param transfrom_country: The path to write the transformed country file to
```
jdddog, line 800:
```suggestion
        :param transform_country: The path to write the transformed country file to
```

oaebu_workflows/workflows/jstor_telescope.py
```diff
@@ -353,304 +361,488 @@ def add_new_dataset_releases(self, releases: List[JstorRelease], **kwargs) -> No
...
-        if report_type is None:
-            logging.info(f"Skipping unrecognized report type, filename {filename}")
-
-        # check format
-        original_email = f"https://mail.google.com/mail/u/0/#all/{message_id}"
-        if extension == "tsv":
-            # add report info
+            save_jsonl_gz(transform_file, results)
+
+    def add_labels(self, reports: List[str]) -> bool:
```
jdddog, line 831:
```suggestion
    def add_labels(self, reports: List[dict]) -> bool:
```

jdddog (changes requested):
Thanks for the changes Keegan, it is looking good. Just a few smaller suggestions.

oaebu_workflows/workflows/jstor_telescope.py
keegansmith21: Nope. Not sure why it was there to begin with but I'll remove it.

jdddog (approved):
Thanks for the changes Keegan

PR #201, opened 2023-11-13
Update irus oapen cloud function version

keegansmith21:


jdddog (approved):
Thanks Keegan, looks good

PR #202, opened 2023-11-22
NOT NULL statements for crossref and GA

keegansmith21:
[JIRA ticket](https://curtinids.atlassian.net/browse/BAD-667)

@kathrynnapier noticed that some entries in the export tables were missing. This appears to be because they're missing an IF NOT NULL statement for crossref and GA.

### DIFF
--- a/oaebu_workflows/database/sql/export_book_author_metrics.sql.jinja2
+++ b/oaebu_workflows/database/sql/export_book_author_metrics.sql.jinja2
@@ -67,7 +67,7 @@ SELECT
         SUM(month.irus_fulcrum.unique_item_requests) as unique_item_requests
     ) as irus_fulcrum,
 FROM `{{ book_product_table_id }}`, UNNEST(onix.authors) as author, UNNEST(months) as month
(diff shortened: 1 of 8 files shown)


kathrynnapier:
@keegansmith21 - I think I may have missed jstor as well, for the author export table anyway. Comparing with the export_book_metrics.sql.jinja2, should array_length be used instead?

OR ARRAY_LENGTH(month.crossref_events) > 0 OR ARRAY_LENGTH(month.jstor_country) > 0

keegansmith21:
> @keegansmith21 - I think I may have missed jstor as well, for the author export table anyway. Comparing with the export_book_metrics.sql.jinja2, should array_length be used instead?
> 
> OR ARRAY_LENGTH(month.crossref_events) > 0 OR ARRAY_LENGTH(month.jstor_country) > 0

@kathrynnapier - Yep jstor condition was also missing for the export tables. I have also changed the conditions to check the array length for crossef and jstor rather than check if they are null.

kathrynnapier (approved):
Thanks @keegansmith21!

PR #203, opened 2023-11-22
File structure change

keegansmith21:
## File structure update

I've updated the file structure for the repository. I used the example from the Academic Observatory Workflows as a template. A brief overview:
- All workflows now have their own directory. This directory contains any schemas, sql and tests that are unique to that workflow.
- Removed the `database` directory and pulled out the `schema` and `sql` folders from it. These contain any schemas/sql that's not specific to a workflow
-  Updated the config.py functions to be able to point to either workflow-specific sql/schemas/fixtures or the global ones
- Removed the `dags` directory as it's not used

### Some notes
- ~~Currently the table schemas are stored in the global `schema` directory. This is because the directories are set up to mirror the bigquery `dataset.table` structure. It would be good to get @jdddog's feedback on whether this is still the best way to structure this.~~
- The `sql` directories are all empty except for the onix_workflow one. This is because it's the only workflow that uses any sql. I have kept the directories anyway as I plan to populate them in the near future.

### New structure snippet:
```
├── oaebu_workflows
│   ├── onix_telescope
│   │   ├── schema
│   │   ├── sql
│   │   └── tests
│   │       └── fixtures
│   ├── onix_workflow
│   │   ├── schema
│   │   ├── sql
│   │   └── tests
│   │       └── fixtures
│   │           ├── e2e_inputs
│   │           ├── e2e_outputs
│   │           └── schema
│   ├── schema
│   │   ├── internet_archive
│   │   └── worldreader
│   ├── sql
│   ├── tests
│   │   └── fixtures
│   │       └── onix_utils
```


jdddog:
Hey @keegansmith21, thanks for restructuring the workflows / sql / schemas like this.

> All workflows now have their own directory. This directory contains any schemas, sql and tests that are unique to that workflow.
> Removed the database directory and pulled out the schema and sql folders from it. These contain any schemas/sql that's not specific to a workflow

I think it would make most sense for the schemas in the oaebu_workflows/schemas folder to go into the schema folder for it's respective workflow module, e.g. schema/onix/onix.json -> onix_telescope/schema/onix.json.

Maybe the top level sql and schema folder can be used for general SQL files and schemas not used by any workflow?

> Removed the dags directory as it's not used
This directory with the init file is still needed for now as it is used when the DAGs are loaded even if there are none in there.

What do you think about removing the workflows module and just putting all of the workflow modules under oaeby_workflows?

PR #204, opened 2023-11-22
include jstor country_name join for collections

kathrynnapier:
The oaebu country reference table has been updated to include a mapping to country names used in jstor usage reports in csv format for collections.

Therefore updated export_book_metrics_country.sql.jinja to join to jstor country_name field to either alpha2 codes or jstor_name. Tested for umich (publisher report with country_name as an alpha2 code, joins to alpha2) and btob (collection with country_name as a string, join to jstor_name).


keegansmith21 (approved):
Thanks for this quick fix Kathryn!

PR #208, opened 2023-12-15
BAD-687

kathrynnapier:
-  link to new GitBook docs page added to index pages and the Background page that is commonly linked to.
-  acknowledgement of GitBook support in README
- links to schema tables will not be updated as we've migrated the documentation to gitbook (schema files are now included in individual telescope folders)


keegansmith21 (approved):
Thanks for making these changes Kathryn!

PR #211, opened 2024-01-24
Updated collections regex

keegansmith21:
JSTOR reports for SHMP are formatted differently to those from BTAA. The regex pattern was missing them. This PR updates the pattern to capture both formats.


jdddog (approved):

PR #212, opened 2024-01-24
astro migration

keegansmith21:
# Astro Migration

This PR refactors the codebase for use with The Astronomer Airflow service. Broadly speaking, the changes are as follows:
- Change DAG creation to use the Airflow Taskflow API
- Upgrade to Ariflow 2.7.3
- Removed setup.py and setup.cfg - oaebu-workflows is no longer an installable package
- Created Dockerfile, which inherits from the Astro airflow image. Used to ship to Astro deployment
- Move all workflow files (and their dependent files) into the dag folder (ready for Astro shipping)
- Move tests to the ./tests folder
- Update the API usage to the new Dataset API (https://github.com/The-Academic-Observatory/observatory-platform/pull/653) that uses BigQuery as a storage service for Release objects
- Update the workflow Release objects to be passed as Xcoms between tasks (a benefit of the taskflow upgrade)
- Collapse the upload tasks into their respective file creation tasks. i.e. upload_downloaded() is absorbed into download(). This is necessary now as we no longer necessarily have persistent disks in Astro deployments.
- Bigquery tasks create the client on behalf of their appropriate publisher, rather than have them all invoke the default environment project.
- Updated telescopes that can have multiple releases to use the new dynamic tasks functionality of the taskflow API


dags/oaebu_workflows/irus_fulcrum_telescope/irus_fulcrum_telescope.py
```diff
@@ -0,0 +1,411 @@
...
+    :param bq_dataset_description: Description for the BigQuery dataset
+    :param bq_table_description: Description for the biguery table
+    :param api_dataset_id: The ID to store the dataset release in the API
+    :param irus_oapen_api_conn_id: Airflow connection ID OAPEN IRUS UK (counter 5)
+    :param catchup: Whether to catchup the DAG or not
+    :param schedule: The schedule interval of the DAG
+    :param start_date: The start date of the DAG
+    """
+    if bq_table_description is None:
+        bq_table_description = "Fulcrum metrics as recorded by the IRUS platform"
```
jdddog, line 139: Could just put directly in the keyword argument?

dags/oaebu_workflows/irus_oapen_telescope/irus_oapen_telescope.py
```diff
@@ -0,0 +1,631 @@
...
+    :param irus_oapen_api_conn_id: The Airflow connection ID for IRUS API - for counter 5
+    :param irus_oapen_login_conn_id: The Airflow connection ID for IRUS API (login) - for counter 4
+    :param catchup: Whether to catchup the DAG or not
+    :param start_date: The start date of the DAG
+    :param schedule: The schedule interval of the DAG
+    :param max_active_runs: The maximum number of concurrent DAG instances
+    """
+
+    if bq_table_description is None:
+        bq_table_description = "OAPEN metrics as recorded by the IRUS platform"
```
jdddog, line 170: Could you put this directly in the keyword argument?

dags/oaebu_workflows/irus_oapen_telescope/irus_oapen_telescope.py
```diff
@@ -0,0 +1,631 @@
...
+        schedule=schedule,
+        start_date=start_date,
+        catchup=catchup,
+        tags=["oaebu"],
+        max_active_runs=max_active_runs,
+        default_args={"retries": 3, "retry_delay": pendulum.duration(minutes=5)},
+    )
+    def irus_oapen():
+        @task()
+        def make_release(**context) -> List[IrusOapenRelease]:
```
jdddog, line 185: Like Google Books, this workflow could also make use of dynamic tasks and a taskgroup for processing releases.

dags/oaebu_workflows/irus_oapen_telescope/irus_oapen_telescope.py
```diff
@@ -0,0 +1,631 @@
...
+
+            releases = [IrusOapenRelease.from_dict(r) for r in releases]
+            for release in releases:
+                # Download files from GCS
+                success = gcs_download_blob(
+                    bucket_name=cloud_workspace.download_bucket,
+                    blob_name=release.download_blob_name,
+                    file_path=release.download_path,
+                )
+                set_task_state(success, context["ti"].task_id, release=release)
```
jdddog, line 326: Maybe better to just add a check and throw a FileNotFoundException if you can't find the file.

dags/oaebu_workflows/jstor_telescope/jstor_telescope.py
```diff
@@ -84,277 +90,325 @@ def __init__(
...
+
+            # Get the reports from GMAIL API
+            api = make_jstor_api(entity_type, entity_id)
+            available_reports = api.list_reports()
+            if not len(available_reports) > 0:
+                raise AirflowSkipException("No reports available. Skipping downstream DAG taks.")
+            return available_reports
+
+        @task
+        def download(available_reports: List[dict], **context) -> List[dict]:
```
jdddog, line 216: Like Google Books, this workflow could make use of dynamic tasks and a task group to process each report.

dags/oaebu_workflows/onix_telescope/onix_telescope.py
```diff
@@ -0,0 +1,322 @@
...
+        dag_id=dag_id,
+        start_date=start_date,
+        schedule=schedule,
+        catchup=catchup,
+        tags=["oaebu"],
+        default_args={"retries": 3, "retry_delay": pendulum.duration(minutes=5)},
+    )
+    def onix_telescope():
+        @task()
+        def make_release(**context) -> List[dict]:
```
jdddog, line 141: Like Google Books, this workflow could make use of dynamic tasks and a task group to process each release.

jdddog (changes requested):
Hey Keegan, nice work on getting the migration done so quickly. See below for some comments on a few things that could be changed.

requirements.txt
```diff
@@ -1,4 +1,13 @@
+# Astro Runtime includes the following pre-installed providers packages: https://docs.astronomer.io/astro/runtime-image-architecture#provider-packages
 beautifulsoup4>=4.9.3,<5
 onixcheck>=0.9.7,<1
```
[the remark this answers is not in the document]
keegansmith21: This would have been really nice, but I'm having trouble getting it to work. I think the issue is that the oaebu dependencies are installed when the astro container is inherited. At this point, git is not installed, so pip can't pull anything using the git+https:// protocol

dags/oaebu_workflows/irus_fulcrum_telescope/irus_fulcrum_telescope.py
keegansmith21: We sure can!

dags/oaebu_workflows/irus_oapen_telescope/irus_oapen_telescope.py
keegansmith21: Sure, would you recommend something like the following:

```python
if not os.path.exists(release.download_path):
    raise FileNotFoundError(f"Could not find file: {release.download_blob_name}")
```

requirements.txt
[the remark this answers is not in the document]
jdddog: Ah I see, that makes sense.

dags/oaebu_workflows/irus_oapen_telescope/irus_oapen_telescope.py
jdddog: Yeah that would be good

jdddog (commented):
Thanks for the changes Keegan, it is looking really good.

Just a few small comments.

We were settings a few parameters in the `DAG` constructor by default that differ to Airflow defaults (unless we override these in a workflow subclass).
  *  `max_active_runs` was set to 1 (Airflow pulls this from airflow.cfg if not specified) 
  * `catchup` was set to False (Airflow defaults this to True)

Could you double check that these are all set to how they were and if not, if the new default settings are OK?

We used to set `"on_failure_callback": on_failure_callback,` in default_args which creates the Slack message on failure, could we set this for each DAG?

Perhaps `retries` and `retry_delay` should be keyword argument parameters in create_dag?

keegansmith21:
> Thanks for the changes Keegan, it is looking really good.
> 
> Just a few small comments.
> 
> We were settings a few parameters in the `DAG` constructor by default that differ to Airflow defaults (unless we override these in a workflow subclass).
> 
> * `max_active_runs` was set to 1 (Airflow pulls this from airflow.cfg if not specified)
> * `catchup` was set to False (Airflow defaults this to True)
> 
> Could you double check that these are all set to how they were and if not, if the new default settings are OK?
> 
> We used to set `"on_failure_callback": on_failure_callback,` in default_args which creates the Slack message on failure, could we set this for each DAG?
> 
> Perhaps `retries` and `retry_delay` should be keyword argument parameters in create_dag?

Hey Jamie, 

Thanks for taking another look at this PR. I've made the following changes:
- Added a max_active_runs=1 default where it was not already set
- Added the on_failure_callback from observatory platform to all dags
- Made the retries and retry_delay parameters arguments for the create_dag function with defaults set to 3 and 5 (minutes) respectively

The catchup was already set for each dag as part of the create_dag defaults.

I also found the default airlfow config args [here](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html) . But it was kind of hard to find so I much prefer explicitly stating the defaults in create_dag.

jdddog (approved):

PR #213, opened 2024-02-05
Added entity_name exception catch

keegansmith21:
JSTOR has altered their report format again. I added a try - except block to catch it.


jdddog (approved):
Thanks Keegan

PR #215, opened 2024-03-18
Remove github actions publish pypi

jdddog:
Remove `.github/workflows/publish-pypi.yml` as this project no longer deploys to PyPI as it is not a Python package.


keegansmith21 (approved):
Thanks for cleaning this up Jamie!

PR #217, opened 2024-03-21
Publisher field

keegansmith21:
- Added ONIX Publishers field to the book product table. Currently contains an array of PublisherName and PublishingRole for each product
- Updated export tables to include the aforementioned PublisherName field (as publisher_name) where the PublishingRole is "Publisher" (same as onix publishingrole code 01)
- Updates the oapen metadata filter to not exclude the publisher (<PublishingDetail> <Publisher>). This means that publisher information will now be supplied by oapen sources


kathrynnapier (approved):
Nice pick up on removing the kibana reference ;)

Thanks very much @keegansmith21, looks good - no other comments from me!

PR #219, opened 2024-04-03
Fix/onix xform file save

keegansmith21:
- Updated filter to properly include published date
- Adds more testing for the filter because things keep sneaking past me
- Fixed bug in onix_utils that looked in the wrong spot for saved file


jdddog (approved):
Thanks Keegan, looks good to me!

PR #220, opened 2024-04-08
Fix/irus sa

keegansmith21:
Updated the irus oapen telescope to use the default service account for its cloud function actions.


jdddog (approved):
Thanks Keegan, looks good

PR #221, opened 2024-04-10
Fix/sensor defaulting

keegansmith21:
Fixes some issues with the sensors. Sensors should now:
- Fail if the most recent target dag run is a fail
- Pass if the most recent target dag run is skipped


jdddog (approved):
Thanks Keegan, looks good

PR #222, opened 2024-04-10
Renamed pool to comply with standard

keegansmith21:
Astro is giving an error with the crossref pool name:

```
    raise InvalidStatsNameException(
airflow.exceptions.InvalidStatsNameException: The stat name (pool.deferred_slots.Crossref Events Pool) has to be composed of ASCII alphabets, numbers, or the underscore, dot, or dash characters.
```

This didn't seem to be an issue before the move, but now the pool isn't editable using the web UI. If the pool isn't working, it may be the cause of the 429 errors we're now receiving from crossref


jdddog (approved):
Good spotting Keegan!

PR #223, opened 2024-04-11
Alternative ONIX snapshot date source

keegansmith21:
This PR adds an optional input to the onix_workflow dag that allows for an alternative onix table to be queried for the purposes of obtaining a snapshot date during make_release. 

The reason for this change is to accommodate the use of views as an alternative to creating redundant onix tables. A view cannot be sharded and thus cannot be queried for its shard date. A workaround to this limitation is to provide the source onix table that the view queries, and instead extract the snapshot date from that.


jdddog (approved):
Thanks for the PR Keegan, looks good to me.

PR #224, opened 2024-04-11
Updated JSTOR collections regex

keegansmith21:
I made this change before the refactor, but somehow it got overwritten/reverted. See https://github.com/The-Academic-Observatory/oaebu-workflows/pull/211


jdddog (approved):

PR #225, opened 2024-04-12
Added exception for different entity format

keegansmith21:
This is another JSTOR PR that got lost in the refactor - see https://github.com/The-Academic-Observatory/oaebu-workflows/pull/213
I also altered one of the test fixtures to hit use both the try and the except


jdddog (approved):
Thanks Keegan!

PR #230, opened 2024-05-10
Fixed publisher name in aggregated queries

keegansmith21:
Some of the export tables were not properly matching the publisher name to the ISBN. This PR fixed this in what is hopefully quite a simple join.


jdddog (commented):
Hey Keegan, the changes look all good to me. The only suggestion is to use spaces rather than tabs for the SQL changes, to make it more consistent with the rest of the codebase.

jdddog (approved):
Thanks for the changes Keegan!

PR #231, opened 2024-05-14
Feature/bespoke data partners

keegansmith21:
SHMP has a lot of historic partners that are implemented in a bespoke way. 

This PR adds functionality to the oaebu_partners to be able to create a DataPartner object at runtime based on an input dictionary (or list of dictionaries). The intention here is to define all of the new data partners in the config.yml file and pass them to the ONIX WF this way, rather than have to hardcode a bunch of new partners into the code.

If this turns out to be a simpler way of passing partners, we may also consider doing with this the original data partners, which are currently _hardcoded_ into the oaebu_partners.py file.


jdddog (approved):
Hey Keegan, thanks for the PR, I'm happy if you merge it. If you think of a way to have less files and still keep things customisable in the future that would be good.

PR #232, opened 2024-05-21
Added clone protection

keegansmith21:
Git clone fails because of an introduced issue in git lfs (originating in git) https://github.com/git-lfs/git-lfs/issues/5749

I've implemented the recommended temporary fix until the issue is resolved.


jdddog (approved):
Thanks for the fix Keegan!

PR #235, opened 2024-05-28
Added product elevation to onix telescope

keegansmith21:
Added the option to elevate the related products when using the classic ONIX telescope. 

Updated tests to ensure this works when turned on.


jdddog (approved):
Thanks for the PR Keegan, the changes look good.

PR #236, opened 2024-05-29
Dataset API and Release updates

keegansmith21:
Updates to how the DatasetAPI and its releases are used. Fixes a bug that was introduced in the refactor that incorrectly labels the 'dataset_id' of the API entries. It should now be much clearer how the API and Release objects interact with bigquery.


jdddog (commented):
Thanks Keegan, the only thing I could suggest is to change the name of the parameter in the DAG from `api_dataset_id` to `api_bq_dataset_id` so that it is easy to see that it is for BigQuery.

jdddog (approved):
Thanks for the changes Keegan, it looks good to me.

PR #237, opened 2024-06-10
SHMP fixes

keegansmith21:
- Added other, other_ltd and other_pay
- Fixed incorrect country matching for muse_country queries


jdddog (approved):

PR #240, opened 2024-06-24
Elevate relation code 13

keegansmith21:
This PR alters the product elevation behaviour to also elevate relation code 13 (Epublication based on print). Previously, the code would also copy all of the related products into elevated products, then replace the original with the parent. This had the effect of making a carbon-copy of the original ISBN with the parent and the related product swapped. This can no longer be done as it only worked logically with OAPEN's metadata, since all of its product relation codes were "06" (alternative format). Instead, we now swap the parent and the child (related product) and invert the relation code (removing any other related product material in the new product). Inverse codes are stated in the [codelist](https://onix-codelists.io/codelist/51). Code 06 is the inverse of itself.

The new method should see no drop in the volume of products elevated, but will see a decline in the new products' related products.

### Old method
- product_1
  - related_product_1.1_code_06
  - related_product_1.2_code_06
  - related_product_1.3_code_13

**Elevates to**
- product_1
  - related_product_1.1_code_06
  - related_product_1.2_code_06
  - related_product_1.3_code_13
- product_1.1
  - related_product_1_code_06
  - related_product_1.2_code_06
- product_1.2
  - related_product_1_code_06
  - related_product_1.1_code_06 

### New method
- product_1
  - related_product_1.1_code_06
  - related_product_1.2_code_06
  - related_product_1.3_code_13

**Elevates to**
- product_1
  - related_product_1.1_code_06
  - related_product_1.2_code_06
  - related_product_1.3_code_13
- product_1.1
  - related_product_1_code_06
- product_1.2
  - related_product_1_code_06
- product_1.3
  - related_product_1_code_27 (27 is the inverse of code 13)

### DIFF
--- a/dags/oaebu_workflows/onix_utils.py
+++ b/dags/oaebu_workflows/onix_utils.py
@@ -537,10 +537,11 @@ def elevate_related_products(onix_products: List[dict]) -> List[dict]:
         if not product_isbn:
             continue  # Only elevate if parent product has an ISBN
 
+        relation_code_map = {"06": "06", "13": "27"}
         for related_product in related_products:
             relation_code = related_product["ProductRelationCode"]
-            if relation_code != "06":
-                continue  # We expect only "Alternative Format" relations - https://onix-codelists.io/codelist/51
+            if relation_code not in relation_code_map.keys():
+                continue  # We expect only "Alternative Format"/Epub relations - https://onix-codelists.io/codelist/51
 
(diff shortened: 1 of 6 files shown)


kathrynnapier (approved):
Thanks @keegansmith21 !

This looks good- the resulting tables show the related products elevate as expected. Aggregation to works and family also looks good.

PR #241, opened 2024-06-26
Subtitle and subjects update

keegansmith21:
This PR makes two changes to the ONIX Workflow outputs:

## Subtitle
The subtitle field is now pulled through from the onix table to the book product and data export tables. Additionally, the title_subtitle field has also been added, which evaluates to `title: subtitle` when both are present, and `title` when there is no subtitle.

## Subjects
Subjects have been incorporated into all of the data export tables now. 
A new `subjects` struct is added to the book product table that includes the subject code, its top-level code and the name of the subject. This struct is propagated through to the data export tables.


jdddog (approved):

PR #242, opened 2024-07-09
Find format from filename, not filepath

keegansmith21:
The date generated by airflow can have a decimal place in it, which makes its way into the filepath. This breaks the regex output that determines the file format. I added a line that extracts the file name in order to determine the format instead.


jdddog (approved):
Thanks Keegan, looks good

PR #243, opened 2024-07-09
Fix/metadata filename regex

keegansmith21:
I tried to do a quick fix and got punished for it :( 

I did it the long way and made a function to extract the file extension. Includes test cases.


kathrynnapier (approved):
Thanks @keegansmith21 - looks good to me.

PR #244, opened 2024-07-16
Onix wf custom schedule

keegansmith21:
Create a custom Airflow timetable for the Onix Workflow. This makes the workflow run every week on Monday and also on the  5th of every month. 

This required the creation of the class, which inherits from Airflow's Timetable base class. We add in the logic for start and end date determination for both scheduled and manual runs. We are also required to register the timetable as a plugin. We do this with the plugin class and also adding the classes to the plugins folder. Weirdly, any DAG that uses the plugin MUST import it relative to the plugin folder. This means that the plugin folder itself must be added to the PYTHONPATH. See [this issue](https://github.com/apache/airflow/discussions/23758#discussioncomment-2769169) for a full explanation. 

Testing this with the sandbox environment required an update to the sandbox's create_dag_run function as it was unable to handle custom timetables. See [this PR](https://github.com/The-Academic-Observatory/observatory-platform/pull/660). Naturally, this means that this PR is dependent on the observatory platform's PR.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -40,6 +40,7 @@ jobs:
       - name: Add dag folder to PYTHONPATH
         run: |
           echo "PYTHONPATH=$HOME/work/oaebu-workflows/oaebu-workflows/dags" >> "$GITHUB_ENV"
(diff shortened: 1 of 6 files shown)


jdddog (approved):

PR #246, opened 2024-07-25
Add subtitle to filter

keegansmith21:
The subtitle field was not being pulled through oapen onix feeds. This is because it was not added to the filter schema so it was being filtered out. This PR adds it to the filter and updates the tests


jdddog (approved):
Thanks Keegan, it looks good to me.

PR #247, opened 2024-08-01
Add subtitle_title to country exoprt SELECT

keegansmith21:
export_country sql was missing the subtitle_title field from the select statement


jdddog (approved):
Thanks Keegan!

PR #248, opened 2024-09-18
Updated schedule for sftp telescopes to everyday

keegansmith21:
The SFTP based telescopes will skip if there are no files in their update folders. It is therefore not very costly to run them everyday instead of once a week. This change is prompted by the update to the ONIX workflow schedule, which now runs more frequently than once per week.


jdddog (approved):
Thanks Keegan, looks good to me

PR #250, opened 2024-10-21
Set trigger rule to none failed

keegansmith21:
Create book task should not trigger unless all upstream tasks did not fail. This PR changes the trigger rule from ALL_DONE to NONE_FAILED


kathrynnapier (approved):
Thanks @keegansmith21 :)

PR #251, opened 2024-10-21
Add filter on SubjectCode to avoid nulls in array

keegansmith21:
The book product SQL cannot handle a null value in any of the subjects field. Apparently we've never encountered this before so it hasn't been an issue until now. I added a secondary WHERE clause to the subquery that creates the array to check if the SubjectCode is NULL. This seems to fix the issue.


kathrynnapier (approved):
Thanks @keegansmith21, it's taken a long time for this new use case to appear!

PR #252, opened 2024-11-06
Error on late GB uploads

keegansmith21:
Since GB requires manual intervention and I'm prone to forget things, I added a check that will return an error if it's past the 4th of the month and the latest partition has not yet been loaded


jdddog (approved):
Thanks Keegan, it looks good to me

PR #253, opened 2025-01-08
Remove partition check

keegansmith21:
I made the incorrect assumption that the sales and traffic tables should always have the same number of partitions. Removing this check.


jdddog (approved):
Thanks Keegan, looks good to me

PR #254, opened 2025-01-08
Work with Row objects, not strings

keegansmith21:
Turns out queries return row objects, not strings. I implemented this change to the early stopping for GB. Updated tests and did a dev run. Seems to work.

Also added a deploy script for development runs


jdddog (approved):

